PR #1, opened 2019-09-06
Initial workflow (OWNERS, nodes, node channels, etc.)

wking:
Still need to work out edge handling, but here's what I have so far for folks to look at.  CC @lucab, @steveeJ


lucab:
Two high level questions on this:
 1. do you plan for Cincinnati to consume all the JSON blobs directly, or do you plan to have an additional aggregation/transformation step in-between? (I'll have further followups based on this)
 2. to avoid confusion, it may be useful to replace the "cincinnati-" prefix in this repo name with something openshift specific. The equivalent of this for Fedora CoreOS (FCOS) is in a repo named "fedora-coreos-streams", under `updates/`

wking:
> do you plan for Cincinnati to consume all the JSON blobs directly, or do you plan to have an additional aggregation/transformation step in-between? (I'll have further followups based on this)

The latter.  A postmerge prow hook will collect all of this, massage it into a single JSON blob, and publish to somewhere outside of GitHub for Cincinnati to consume (S3?  An app-sre ConfigMap?  Quay labels?).

> to avoid confusion, it may be useful to replace the "cincinnati-" prefix in this repo name with something openshift specific.

The org name is OpenShift-specific, and folks who need to distinguish between multiple Cincinnati can rename of fork (e.g. my fork of openshift/release is wking/openshift-release).

But on both points, the repo is new.  Push back if I'm not convincing ;).

lucab:
All good.

On the former, it's nice to have a massaged format. It'd be good to keep that closer to the way we query the key-value quay API.
(Perhaps you can even think of making this format human-friendlier if people are supposed to write entries by hand, like YAML or TOML)

On the latter, it's just a minor naming concern. Cincinnati is originally the name of the protocol replacing Omaha. We are slowly conflating implementation AND metadata format under that name now. But only specifically for openshift API. However, there are already other implementations of the protocol (e.g. the endpoint for nightlies) and other metadata formats (e.g. the FCOS one).

wking:
> Perhaps you can even think of making this format human-friendlier if people are supposed to write entries by hand, like YAML or TOML

JSON looks like Python dicts and JavaScript objects, so it can't be too human-unfriendly ;).  But sure, as long as a YAML/TOML parser is an acceptable dependency for the massaging/CI tools, I'd be ok with that.  I'm fine deferring to the release managers; it should be easy to change later.

> However, there are already other implementations of the protocol (e.g. the endpoint for nightlies)...

My long-term goal is to get nightlies into this flow and serve a graph for `nightly-4.1`, etc., out of this repo and the production OpenShift Cincinnati.

> ... and other metadata formats (e.g. the FCOS one).

I'd expect FCOS metadata to flow out of repos under other orgs.  But no sense in diverging data formats.  I'll take a closer look at the format in that repo.

wking:
Checking [the FCOS node definition][1], it looks like most of the time is spent distinguishing between multiple arches and formats.  These rolls will eventually be managed in OpenShift with multi-arch images, so I don't think we need to be quite so elaborate, especially out of the gate when we are amd64-only.  Do you think the FCOS feeder and this Cincinnati feeder are consuming different-enough artifacts to motivate different input formats?  Or do you think my current `nodes/` structure should be adjusted somehow?

[1]: https://github.com/coreos/fedora-coreos-streams/blob/master/streams/testing.json

wking:
I've pushed up bbcf2d8 adding edge handling and doing some other polishing.

README.md
```diff
@@ -0,0 +1,47 @@
...
+* `payload` (required string): The by-digest pullspec for the release [container image][image].
+* `version` (required string): The name of the release.
+    Currently [a Semantic Versioning 2.0.0][SemVer] string.
+
+For an example, see [`nodes/4.1/4.1.0.json`](nodes/4.1/4.1.0.json).
+
+### Add a release to a channel
+
+Add a symlink to the `nodes` entry from the `channels/<channel>` directory.
+For example, [`channels/stable-4.1/4.1.0.json](channels/stable-4.1/4.1.0.json) is a symlink to [`nodes/4.1/4.1.0.json`](nodes/4.1/4.1.0.json), which promotes the 4.1.0 release to the `stable-4.1` channel.
```
crawford:
```
[`channels/stable-4.1/4.1.0.json`]
```

edges/4.1/4.1.0-rc.6/4.1.0-rc.5.json
```diff
@@ -0,0 +1,7 @@
+{
+  "channels": [
```
crawford: Why are edges associated with channels?

README.md
wking: >    [`channels/stable-4.1/4.1.0.json`]

Fixed with 5b6281e -> bfd6034.

edges/4.1/4.1.0-rc.6/4.1.0-rc.5.json
wking: > Why are edges associated with channels?

Setting up for:

```console
$ cat edges/4.1/4.1.15/4.1.14-prerelease.json 
{
  "channels": [
    "prerelease-4.1"
  ],
  "from": "4.1.14",
  "to": "4.1.15",
  "metadata": {
    "start": "2019-09-16T20:00Z",
    "duration": 360,
  }
}
$ cat edges/4.1/4.1.15/4.1.14-stable.json 
{
  "channels": [
    "stable-4.1"
  ],
  "from": "4.1.14",
  "to": "4.1.15",
  "metadata": {
    "start": "2019-09-17T20:00Z",
    "duration": 360,
  }
}
```

or similar to stage the stable edge rollout a day after the prerelease rollout (cf. https://github.com/coreos/fedora-coreos-streams/pull/10).  And sharing metadata about start dates here so we don't have to have multiple Cincinnati pods communicating with each other about when they decided to start the rollout.

crawford: Hmm. I guess I would expect the channels to be associated with nodes instead. Can the channel be specified on both edges and nodes? What happens if they differ?

wking: It will be a CI error in this repo to define an edge in a channel where either node does not belong to that channel.

wking:
I've pushed a WIP script for publishing this data to Quay labels, which we can use until we work out a better way to feed it into Cincinnati.  Doesn't actually have POST/DELETE support yet, but you can run it to see what it would change:

```console
$ hack/push-to-quay.py 
changing 4.1.0-rc.6 previous.add from [] to [u'4.1.0-rc.5']
(u'4.1.0-rc.6', 'FIXME post', 'https://quay.io/api/v1/repository/openshift-release-dev/ocp-release/manifest/sha256:1dabe42b5c94841fd8736d8f3a80afeaf5f5ad3833cef8d304c419a97b0efbc3/labels')
(u'4.1.0-rc.5', 'FIXME delete', 'https://quay.io/api/v1/repository/openshift-release-dev/ocp-release/manifest/sha256:dc67ad5edd91ca48402309fe0629593e5ae3333435ef8d0bc52c2b62ca725021/labels/io.openshift.upgrades.graph.next.add')
changing 4.1.0-rc.8 previous.add from [u'4.1.0-rc.7'] to []
(u'4.1.0-rc.8', 'FIXME delete', 'https://quay.io/api/v1/repository/openshift-release-dev/ocp-release/manifest/sha256:8250bbf79d4f567e24a7b85795aba97f9e75a9df18738891a0cb6ba42e422584/labels/io.openshift.upgrades.graph.previous.add')
```

That's replacing the `next.add` 4.1.0-rc.5 -> 4.1.0-rc.6 with a `previous.add` and removing the redundant `4.1.0-rc.7`->`4.1.0-rc.8` label, which makes sense.

wking:
Ok, fixed up the POST and DELETE handling (although I don't have creds to actually run those, so there are a handful of untested lines in the script.  That should be all we need to start using this repo to manage the production graph metadata, vs. directly editing Quay tags.  I'll start a follow-up branch to add CI checks to catch things like "listed version does not exist", [impossible edge][1], cyclic graph, version removed from channel, etc.  And once this branch lands, I'll add a postsubmit job to Prow to automatically publish after a PR merges.

[1]: https://github.com/openshift/cincinnati-graph-data/pull/1#discussion_r321950377

lucab:
@wking the metadata you were looking at for FCOS is only used for boot-images. The relevant ones for Cincinnati are (example from the `testing` stream):
 * [release-index](https://builds.coreos.fedoraproject.org/prod/streams/testing/releases.json): describes the set of **nodes** in the graph
 * [updates-metadata](https://builds.coreos.fedoraproject.org/updates/testing.json): describes (cacheable) **edges** in the graph

>  Do you think the FCOS feeder and this Cincinnati feeder are consuming different-enough artifacts to motivate different input formats? Or do you think my current nodes/ structure should be adjusted somehow?

I think it is fine if they diverge, just let's try to keep the overall approach similar. Relevant semantic differences in FCOS are:
 * there is a graph **per stream**. Clients cannot automatically jump across streams via auto-updates
 * there is a single graph for all architectures in the stream. That is, update paths are not architecture dependent. Some nodes may be missing for a specific arch, though.

wking:
> * updates-metadata: describes (cacheable) edges in the graph

How does this represent edges?  It has a `releases` array, but each entry in that array only mentions one version.  Is that the `to` version?  What is the `from` version?

lucab:
Correct, a rollout only specifies the `to` version.
The `from` set is composed of all the releases that come before it (in the release-index array) and are not deadends.
(There is another special case for enforcing barriers/chokepoints, but we haven't finalized/tested it yet).

You can observe the current FCOS `testing` graph here:
```
curl -H 'Accept: application/json' 'https://updates.coreos.stg.fedoraproject.org/v1/graph?basearch=x86_64&stream=testing&rollout_wariness=0'
```

Note: all of these are moving pieces as of today, full details will be in a public repo once settled.

lucab:
Additionally, we just triggered a 24h-spread rollout right now, procedure is https://github.com/coreos/fedora-coreos-streams/pull/17.

wking:
> The `from` set is composed of all the releases that come before it (in the release-index array) and are not deadends.

Say a user upgrades from A->B, and then some flaw is discovered in B and you mark it a dead-end.  How do they get off?  With explicit from/to you can tune this sort of thing directly, and tooling can automate the tedium (e.g. see 1dafd8c89b300f898da20ecddc93828edfbcd439).

lucab:
By design, a dead-end cannot auto-update further (e.g. because the update logic is broken, or some major breaking change happened somewhere in the OS) and requires manual intervention.

All the other cases would be handled by one (or more) outgoing edges from B, by defining one (or more) in progress rollouts.

I do acknowledge that you can make releng tooling as complex as you want, assuming you have enough knowledge/manpower on the releng team.
FCOS however is a community project with very limited resources, so the goal of this design is to mimic the ContainerLinux flow while introducing minimal complexity.

wking:
Ok, so I'm comfortable with diverging in-repo schemas, because I think OpenShift need a way to suggest offramps from bad releases, and I think we want the ability to limit upgrade recommendations to edges we have tested (e.g. no 4.1.0->4.1.15 in 1dafd8c).  Are you comfortable with my current schema here?

edges/4.1/4.1.0-rc.6/4.1.0-rc.5.json
crawford: I see. So this is an edge whitelist rather than our current blacklist approach.

wking: > So this is an edge whitelist rather than our current blacklist approach.

Yeah, with this approach we'll scrape baked-in edges out of the release image here, rather than requiring Cincinnati to extract those directly.  The push-to-quay approach is just a cludge until we have a Cincinnati plugin to slurp some yet-to-be-determined JSON output from this repo.

lucab:
Yes, fine to diverge, and I trust your judgment as I am not on-top of OpenShift requirements/usecases.

I only share @crawford concern wrt. attaching metadata to edges (as proposed here) instead of nodes because:
 * Cincinnati graph does not have annotated edges (today)
 * annotated nodes are closer to the quay-metadata model and can be easily ported to any K-V store mechanism (making the design easier to reason about)

crawford:
I'm concerned at the amount of duplication between the release-metadata from release images and this repo. I'm sure that we _can_ have CI ensure that they don't differ, but I'm not convinced we should create the work for ourselves in the first place. How do you envision the definition of each node being used? Do we need anything more than a tag or digest?

I'm also slightly uneasy about the level of verbosity. I'm okay with the edge whitelist approach, but I think I'd prefer a more condensed approach.

Lastly, as @lucab mentioned, if we expect humans to craft these payloads, they can't be JSON.

wking:
> I'm concerned at the amount of duplication between the release-metadata from release images and this repo.

I'm trying to set this repo up as a middleman so Cincinnati doesn't have to check `release-metadata` directly.  Plus, once we start phasing in all edges, we'll need the per-edge metadata here.

> I'm sure that we can have CI ensure that they don't differ, but I'm not convinced we should create the work for ourselves in the first place.

Shouldn't be too much work.  1dafd8c was a two-liner (see the commit message).

> How do you envision the definition of each node being used? Do we need anything more than a tag or digest?

Tag, digest, and errata URI are all we need now to keep Cincinnati out of `release-metadata`.

> I'm also slightly uneasy about the level of verbosity. I'm okay with the edge whitelist approach, but I think I'd prefer a more condensed approach.

Condensing (with the FCOS schema as an extreme example) makes per-edge-channel manipulation trickier.  For example, say B proves unstable and we want to pull A-B and recommend B->A.  

> Lastly, as @lucab mentioned, if we expect humans to craft these payloads, they can't be JSON.

I'll convert to YAML.

wking:
I've pushed f0d155a with a new `graph-util.py` subcommand to support:

```console
$ hack/graph-util.py update-node quay.io/openshift-release-dev/ocp-release:4.1.0
```

to seed the node JSON without having to actually touch JSON.  That means a full "take a new release image into prerelease-4.1 with its baked-in edges would look like:

```console
$ hack/graph-util.py update-node quay.io/openshift-release-dev/ocp-release:4.1.15
$ ln -rs nodes/4.1/4.1.15.json channels/prerelease-4.1/
$ hack/graph-util.py extract-edges 4.1.15
```

And then promoting to stable-4.1 would be:

```console
$ ln -rs nodes/4.1/4.1.15.json channels/stable-4.1/
$ hack/graph-util.py extract-edges 4.1.15
```

PR #2, opened 2019-09-12
WIP: Initial workflow (OWNERS, etc.) only tracking secondary metadata

wking:
While #1 is currently only updating Quay lables, there has been some concern around it positioning itself to be the single source of Cincinnati graph-builder data at some future point.  Since my main goal is to get the release-admin workflow into version control, here's a lighter-weight alternative that only manages secondary metadata.  Cincinnati would continue to pull primary metadata dynamically from image `release-metadata` on Quay, and it would get overrides from this repository.

For at least the short term, those overrides would be set in Quay labels just like they are today; we might change that later, but it wouldn't affect the release-admin workflow.

There's also been pushback against hand-edited JSON in #1.  But I'm lazy and want these scripts to work even for folks without PyYAML installed ;).  We can switch to YAML easily later if folks are comfortable with the workflow.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,61 @@
+# How to Contribute
+
+This project is [Apache 2.0 licensed](LICENSE) and accept contributions via GitHub pull requests.
+This document outlines some of the conventions on development workflow, commit message formatting, contact points and other resources to make it easier to get your contribution accepted.
+
+## Security response
+
+If you've found a security issue that you'd like to disclose confidentially, please contact Red Hat's Product Security team.
+
+## Getting started
+
+- Fork the repository on GitHub.
+- Read the [README](README.md) for build and test instructions.
+- Play with the project, submit bugs, submit patches!
+
+### Contribution flow
+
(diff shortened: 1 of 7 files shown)


steveeJ:
Thanks @wking, this is much like what I originally had in mind. Ideally I would've liked if pushing the tags could've happened from CI which we have deemed to risky for now in  an out-of-band discussion.

Which brings me to the most interesting part of the validation rules when to enforce them.
We can't rule out that someone accidentally pushes the tags without going through CI and review first, so we have to enforce at push-time.
The one rule we have discussed so far, which IIUC has lead to this PR is:
* `graph.release.channels` must be append only, as we never want a release removed from a channel.
We might want an override flag which takes the rule-name for exceptional cases.

We have a quay v1 API client over at [cincinnati](https://github.com/openshift/cincinnati/tree/276c848f9c947f6477cf91692e07eff6020cbc6b/quay). It also has support for token authentication and fetching the labels. It'd be great to extend that with pushing labels.

wking:
replaced by #4.

/close

PR #3, opened 2019-12-02
Add a description of the objects we talked about

eparis:


objects.md
```diff
@@ -0,0 +1,118 @@
+# About
+
+This repo needs to represent the following information:
+
+1. Adding/removing versions to channels
```
wking: "Adding releases to channels"?  We never remove them from channels to avoid stranding users who need off ramps.

objects.md
```diff
@@ -0,0 +1,118 @@
...
+2. Remove or add edges to all channels
```
wking: This happens at release-build time, because we want, but don't currently require, edge recommendations to be part of the signed release content (to guard against compromised Cincinnati).  Also, once we grow phased edge roll-outs, we're going to want per-channel start times and durations for those roll-outs (see also [here][1]).

[1]: https://github.com/openshift/cincinnati-graph-data/pull/1#discussion_r321911263

objects.md
```diff
@@ -0,0 +1,118 @@
...
+3. Remove or add edges to a specific channel
```
wking: not sure why we'd ever remove from a single channel.  Do you have a use-case for that?

objects.md
```diff
@@ -0,0 +1,118 @@
...
+4. Implement an "upgrade policy" object specifically for y-streams
+
+Remember that cininnatti should pull most edges from the release manifest. This data should typiclly remove edges, not add them.
```
wking: nit: "cincinnati" -> "Cincinnati" and "typiclly" -> "typically"

objects.md
```diff
@@ -0,0 +1,118 @@
...
+- /added_edges/
+- /blocked_edges/
+- /y-policies/
+
+# Nodes
+
+The /nodes objects are very simple. It is not strictly necessary as long as cincinnatti pulls all releases from quay. However it is useful for a CI tool to build a final (smaller) graph and for disconnected customers to build a complete cincinnatti graph. (though the later is not the first target goal!) So /nodes/4.1.24 would look like:
+```
+version: 4.1.24
+payload: quay.io/openshift-release-dev/ocp-release@sha256:6f87fb66dfa907db03981e69474ea3069deab66358c18d965f6331bd727ff23f
```
wking: This saves us a tag-listing query, but we still have to hit the registry to get metadata, etc.  Doesn't seem to be worth the trouble.  I'd drop it unless it it grows to look like #1 and include all node metadata (so Cincinnati doesn't have to hit the registry at all).

objects.md
```diff
@@ -0,0 +1,118 @@
...
+The /channels object described the nodes in a channel. For example the /channels/stable-4.1.yaml file would contain
+
+```
+name: stable-4.1
+versions:
+  - 4.1.0
+  - 4.1.1
+  - ...
+  - 4.1.11
+  # 4.1.12 Was excluded from stable!!!
```
wking: I don't think we want comments like this.  The set of excluded nodes can be calculated by comparing available nodes with the explicitly added nodes and noticing that (for example) 4.1.12 was not in stable-4.1.  If we want to record information about explicit exclusion decisions, I'd rather close the "add 4.112 to stable-4.1" PR with a message about why we were rejecting the release, instead of landing a comment about why we were rejecting the release.

objects.md
```diff
@@ -0,0 +1,118 @@
...
+  - 4.1.14
+subchannels:
+```
+
+Subchannels can be implemented later. The idea is that inside stable-4.2.yaml we would have:
+```
+subchannels:
+  - stable-4.1
+```
+Which would pull all nodes from stable-4.1 into stable-4.2. Without having to explicitly add them.
```
wking: Can we drop this business about subchannels?  I think we want to restrict `stable-4.2` tagging to 4.1 releases that we (at one point) felt were good jumping-off points for 4.1->4.2 updates.  I don't see a need to tag older 4.1 releases into 4.2 and would prefer other approaches for "I'm on some ancient 4.1.z and want to get over to 4.2".  But we can sort that all out somewhere else without blocking a Git flow for our current models if we drop the subchannel references here.

objects.md
```diff
@@ -0,0 +1,118 @@
...
+
+# Added Edges
+
+Adding edges will be rare. It will have the same format as removed edges. It will typiclly be used only in conjunction with 'policy' below.
+
+# Blocked Edges
+
+Blocked edges will be used any time we find a problem. The blocked edge object for /blocked_edges/4.1.10.yaml would looke like:
```
wking: nit: "looke" -> "look"

objects.md
```diff
@@ -0,0 +1,118 @@
...
+```
+to: 4.1.10
+- channels:
+    - *
```
wking: I don't see a reason to have channel-specific blocking.  Can we drop this unless we decide we need it?

objects.md
```diff
@@ -0,0 +1,118 @@
...
+  from: *
```
wking: Definitely need this, e.g. to exclude 4.1.z->4.2.2 without excluding 4.2.0->4.2.2.  But I'd rather use a regexp or explicit list instead of this glob (or maybe it's a single special character, and not a glob?).

objects.md
```diff
@@ -0,0 +1,118 @@
...
+- channels:
+    - stable-4.1
+    - prerelease-4.1
+  from: 4.1.7
+......
+```
+
+# Y-Policies
+
+Policy is a very OpenShift y-stream specific edge auto-snipping during y-stream transitions. We need to build payloads with all possible edges and have those snipped while we determine what is safe to give to customers. We would like this 'auto snipping' policy to be encoded in this repo. We think that /y-policies/4.3.yaml would looke like:
```
wking: nit: I'd rather stick to "remove edge" instead of adding a new "snip" verb.

objects.md
eparis: yes. when we went from rhel 8 to rhel 8.1 we were happy to put the edge in the candidate channel, but not in the stable channel. While we still wanted to have both of the nodes in the stable channel.  Eventually we wanted to put the edge in stable, but we wanted to be able to blacklist it for a while.

objects.md
eparis: agree that I don't currently know a reason to 'remove'

objects.md
eparis: adding happens at build time. yes.

however with the 'policy' object we would really like, we need some way to 're-add' back to what was in payload.

objects.md
eparis: I believe that for 'disconnected' people this format would allow someone to generate a complete graph (even without the in payload metadata)

I completely agree, not strictly required to start.

objects.md
eparis: Comments like this should/would come from architects to make it 'easy' for a human to know that a version missing is explicit. I believe such comments are absolutely non-functional non-required, but we agreed we would use them if we were still doing this 'by hand'

objects.md
eparis: agree that subchannels are not required in an initial implementation. But I'm not convinced to give up on the idea...

objects.md
eparis: I was thinking of `*` as a single special character, not a glob.  If we need anything other than a single character I'm fine with a regex, not glob.

objects.md
eparis: See the comment about the update when we switched from RHEL8.0 to 8.1. We wanted per-channel edge control. It can't be accomplished using quay labels. So I wouldn't expect it to necessarily work at first.

wking:
I have [a branch up][1] with a script implementing parts of this.  Current output (as of e73243a1bdd) looks like:

```console
DEBUG: loaded from cache: 4.2.10 quay.io/openshift-release-dev/ocp-release@sha256:dc2e38fb00085d6b7f722475f8b7b758a0cb3a02ba42d9acf8a8298a6d510d9c
DEBUG: loaded from cache: 4.2.10-s390x quay.io/openshift-release-dev/ocp-release@sha256:28e9bbf8aa8c97e7bff1eeb2666ed6bde0edd8a72fc9b595fd1f5a367bf65a76
...
DEBUG: loaded from cache: 4.0.0-5 quay.io/openshift-release-dev/ocp-release@sha256:ed45d79377a8f36aa5db7be63c1278ae864031ec3d50b80fdd9902c6be4517ba
WARNING: the io.openshift.upgrades.graph.next.add label is deprecated.  Use the previous label on the other release(s) instead (was: 4.1.0-rc.6)
INFO: 4.1.0-rc.5 delete https://quay.io/api/v1/repository/openshift-release-dev/ocp-release/manifest/sha256:dc67ad5edd91ca48402309fe0629593e5ae3333435ef8d0bc52c2b62ca725021/labels/io.openshift.upgrades.graph.next.add
INFO: label sort for 4.1.18: prerelease-4.1,stable-4.1,candidate-4.2 -> candidate-4.2,prerelease-4.1,stable-4.1
INFO: 4.1.18 delete https://quay.io/api/v1/repository/openshift-release-dev/ocp-release/manifest/sha256:420633acf3fc7572372fe2df758152f6ab1f53a21c79a6c4b741fa0394c7df3a/labels/io.openshift.upgrades.graph.release.channels
INFO: 4.1.18 post https://quay.io/api/v1/repository/openshift-release-dev/ocp-release/manifest/sha256:420633acf3fc7572372fe2df758152f6ab1f53a21c79a6c4b741fa0394c7df3a/labels
...
INFO: replacing 4.2.0-rc.5 previous remove u'4.1.18,4.1.20' with *
INFO: 4.2.0-rc.5 delete https://quay.io/api/v1/repository/openshift-release-dev/ocp-release/manifest/sha256:1a2b56060f9d462ce879ce57d27f41941081ba063f4ac7f775505f0903d0713a/labels/io.openshift.upgrades.graph.previous.remove
INFO: 4.2.0-rc.5 post https://quay.io/api/v1/repository/openshift-release-dev/ocp-release/manifest/sha256:1a2b56060f9d462ce879ce57d27f41941081ba063f4ac7f775505f0903d0713a/labels
```

The `next.add` label on 4.1.0-rc.5 should be a `previous.add` label on 4.1.0-rc.6. There's no need to use the `next` labels.  The `label sort` changes are just for consistent ordering (where the currently-live labels are in whatever-the-admin-felt-like ordering).  The 4.2.0-rc.5 change from `4.1.18,4.1.20` to `*` is the same effect, but says "there are no incoming edges" in a way that's more robust than a by-name blacklist.  Anyhow, checkout my branch, kick the tires, and incorporate it here (or let me know if you want me to carry in a new PR), or let me know what still needs changing.

[1]: https://github.com/wking/cincinnati-graph-data/tree/eparis

wking:
replaced by #4.

/close

PR #4, opened 2020-01-03
Declarative method to sync labels to quay

eparis:


wking:
/lgtm

Let's just land this and iterate later :)

wking:
Oh, maybe no Tide merges because openshift/release#4958 never landed.

PR #8, opened 2020-01-03
channels: promote 4.1.29 and 4.2.13

crawford:


channels/candidate-4.2.yaml
```diff
@@ -27,3 +27,5 @@ versions:
 - 4.2.11-s390x
 - 4.2.12
 - 4.2.12-s390x
+- 4.2.13
+- 4.2.13-s390x
```
wking: you want 4.1.29 in this file too, right?  Up^^^ with the other 4.1.z to allow folks to make the 4.1->4.2 hop

wking: Done in da0fd3b -> 2b3bf1c

PR #9, opened 2020-01-03
Revert "Process versions in semver order"

wking:
This reverts commit 830db32ea5e358f81f6f38e32802dd50447e5727, #4.

Sorting by string is stable, and not terribly far from SemVer sorting.  Exact SemVer sorting would be nice, but is not important enough to be worth adding a dependency outside Python's standard libraries.

### DIFF
--- a/hack/graph-util.py
+++ b/hack/graph-util.py
@@ -8,7 +8,6 @@ import logging
 import os
 import re
 import tarfile
-import semantic_version
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


wking:
/assign @eparis 

since I'm reverting your commit ;)

eparis:
since i'm going to make it multithreaded i'm fine with this.  It originally was valuable for me to see, in the printout, nicely ordered list of channels

PR #11, opened 2020-01-03
Make the calls to sync multithreaded

eparis:
the quay API is slow.

With this change:
time hack/graph-util.py push-to-quay
real	0m15.458s

Without this change:
time hack/graph-util.py push-to-quay
real	2m45.408s

### DIFF
--- a/hack/graph-util.py
+++ b/hack/graph-util.py
@@ -2,15 +2,16 @@
 
 import argparse
 import codecs
+import functools
 import io
 import json
 import logging
+import multiprocessing
 import os
 import re
 import tarfile
 
 import yaml
-
 
 try:
     from builtins import FileExistsError  # Python 3
@@ -166,10 +167,17 @@ def push(directory, token, push_versions):
     nodes = load_nodes(directory=os.path.join(directory, '.nodes'), registry='quay.io', repository='openshift-release-dev/ocp-release')
     nodes = load_channels(directory=os.path.join(directory, 'channels'), nodes=nodes)
     nodes = block_edges(directory=os.path.join(directory, 'blocked-edges'), nodes=nodes)
-    for version, node in sorted(nodes.items()):
-        if push_versions and version not in push_versions.split(','):
-            continue
-        sync_node(node=node, token=token)
+
+    sync_nodes = [
+        node for version, node in sorted(nodes.items())
+        if not push_versions or version in push_versions.split(',')
+    ]
+
+    sync = functools.partial(sync_node, token=token)
+    pool = multiprocessing.Pool(processes=16)
+    pool.map(sync, sync_nodes)
+    pool.close()  # no context manager with-statement because in Python 2: AttributeError: '__exit__'
+
 
 def update_channels(node, token):
     labels = get_labels(node=node)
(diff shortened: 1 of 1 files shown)


hack/graph-util.py
```diff
@@ -11,17 +11,9 @@
...
-
-try:
-    from builtins import FileExistsError  # Python 3
-except ImportError:
-    FileExistsError = OSError  # sloppy hack for Python 2
-
-try:
-    from urllib.request import Request, urlopen  # Python 3
-except ImportError:
-    from urllib2 import Request, urlopen  # Python 2
```
wking: Do we need to drop Python 2 support?  Looks like [they have `Pool` too][1], it's just not a context manager:

```console
$ python2
Python 2.7.5 (default, May 31 2018, 09:41:32) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-28)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import multiprocessing
>>> with multiprocessing.Pool(processes=16) as pool:
...     print(pool)
... 
  File "<stdin>", line 1, in <module>
AttributeError: __exit__
```

[1]: https://docs.python.org/2/library/multiprocessing.html#multiprocessing.pool.multiprocessing.Pool

wking: ah, maybe [`starmap`][1] is new in 3.3.

[1]: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.starmap

hack/graph-util.py
```diff
@@ -166,10 +158,15 @@ def push(directory, token, push_versions):
     nodes = load_nodes(directory=os.path.join(directory, '.nodes'), registry='quay.io', repository='openshift-release-dev/ocp-release')
     nodes = load_channels(directory=os.path.join(directory, 'channels'), nodes=nodes)
     nodes = block_edges(directory=os.path.join(directory, 'blocked-edges'), nodes=nodes)
+
+    sync_nodes = []
     for version, node in sorted(nodes.items()):
         if push_versions and version not in push_versions.split(','):
             continue
-        sync_node(node=node, token=token)
+        sync_nodes.append((node, token))
```
wking: no need to append the token here, you can do something like:

```python
sync_nodes = [
    node for version, node in sorted(nodes.items())
    if not push_versions or version in push_versions.split(',')
]
with Pool(processes=16) as pool:
    pool.starmap(lambda node: sync_node(node=node, token=token), sync_nodes)
```

hack/graph-util.py
wking: but we may be fine with 2.7's `map` and not need `starmap`...

wking:
I've pushed [this branch][1] with 3c25e3be495 to recover Python 2 support ;)

[1]: https://github.com/wking/cincinnati-graph-data/tree/python2-multithread-fixup

wking:
Also, multiple processes are a pretty big sledgehammer for parallelizing network requests.  But [from Python's docs][1]:

> The multiprocessing module also introduces APIs which do not have analogs in the threading module. A prime example of this is the Pool object...

This script is too peripheral to be worth building thread-based pool, and we can always pivot to some lighter weight parallel harness if someone finds a better one in the stdlib ;).

[1]: https://docs.python.org/3/library/multiprocessing.html#introduction

hack/graph-util.py
eparis: I tried to use a closure, and those don't pickle and so won't work with map. I'll try with a lambda and see how if works.

eparis:
```
AttributeError: Can't pickle local object 'push.<locals>.<lambda>'
```

eparis:
you changes work for me! repushed!

eparis:
I'm also happy to use 
```
from multiprocessing.dummy import Pool as ThreadPool
```
So it uses threads instead of processes. It seems to work in both python and that old python2 junk   :)

wking:
> ```python
> from multiprocessing.dummy import Pool as ThreadPool
> ```

Oooh :).  Can we `import multiprocessing.dummy` and then have `pool = multiprocessing.dummy.Pool(...)`?  Then readers don't have to hunt down the `import` line to know where the pool class is coming from.

PR #12, opened 2020-01-06
channels: promote 4.1.29 and 4.2.13

crawford:


vrutkovs:
Looks good, not sure if these need to be removed for `prerelease` and `candidate` channels

crawford:
@vrutkovs We leave the versions in prerelease and candidate so that clusters which follow those channels can still successfully upgrade.

PR #14, opened 2020-01-09
add 4.2.14 in candidate-4.2

sosiouxme:
Note that the s390x release is also named `4.2.14`. I'm curious if this magically works to include both.


wking:
> Note that the s390x release is also named 4.2.14. I'm curious if this magically works to include both.

Yes, we want arch-agnostic names.  Cincinnati handles arch-specific sorting internally since openshift/cincinnati#180.

channels/candidate-4.2.yaml
```diff
@@ -30,3 +31,4 @@ versions:
 - 4.2.12-s390x
 - 4.2.13
 - 4.2.13-s390x
+- 4.2.14
```
wking: oh, wait, because both the s390x and amd64 images share this name, we want `4.2.14+amd64` here to only promote the amd64 release.  Or 4.2.14 to promote both amd64 and s390x.  I'll work up a patch to the pusher script to handle the optional build information.

sosiouxme: cool, that was gonna be my question... what if we decide not to promote one (though frankly that would be a mess for other reasons... we can't separate out the releases in errata-tool or in the appregistry)

wking: #16 has my WIP in this direction (just the README, but folks can discuss my approach there while I work up the implementation).

wking: and this PR is good for promoting both as it stands (assuming my approach holds up ;).

And you could promote both to candidate channels but then only promote amd64 to fast/stable if, for example, s390x turned out to be unstable.  Does that address the errata concern?

wking:
We want https://bugzilla.redhat.com/show_bug.cgi?id=1789765 to get sorted before we land this

wking:
We may want to exclude s390x until we have a handle on https://bugzilla.redhat.com/show_bug.cgi?id=1789260

PR #16, opened 2020-01-10
hack/graph-util: Handle architecture build metadata

wking:
So folks can distinguish between the 4.2.14 build for amd64 and the one built for s390x, etc.

```console
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.2.14-x86_64 | grep 'Name\|Digest\|Arch'
Name:          4.2.14
Digest:        sha256:3fabe939da31f9a31f509251b9f73d321e367aba2d09ff392c2f452f6433a95a
OS/Arch:       linux/amd64
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.2.14-s390x | grep 'Name\|Digest\|Arch'
Name:          4.2.14
Digest:        sha256:26ba0c17618872c68a2e9a5002c3f3e48f75ae844fcf66e83063bd72a104cfc7
OS/Arch:       linux/s390x
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,14 @@ The [contributing documentation](CONTRIBUTING.md) covers licencing and the usual
 
 **Do Not Ever Update Quay Labels Based On Your Local Development Branch. Only from master!**
 
+### Release names
+
+Release names are used for [adding releases to channels](#add-releases-to-channels) and [blocking edges](#block-edges).
+Architecture-agnostic names will apply to all images with that exact name in the `version` property of the `release-metadata` file included in the release image.
+Names with [SemVer build metadata][semver-build] will apply only to releases whose exact name in the `version` property matches the release with the build metadata removed and whose referenced [image architecture][image-arch] matches the given build metadata.
+For example, 4.2.14 will apply to both the amd64 and s390x release images, since those both have `4.2.14` in `version`.
+And 4.2.14+amd64 would only apply to the amd64 release image.
+
 ### Add Releases To Channels
 
 Edit the appropriate file in `channels/`.
@@ -54,3 +62,5 @@ $ hack/graph-util.py push-to-quay --token="${YOUR_TOKEN}"
 You can leave `--token` unset for a dry run (the actions the script would take are printed either way, but are only executed if you passed a token).
 
 [Cincinnati]: https://github.com/openshift/cincinnati/
+[image-arch]: https://github.com/opencontainers/image-spec/blame/v1.0.1/config.md#L103
+[semver-build]: https://semver.org/spec/v2.0.0.html#spec-item-10
--- a/hack/graph-util.py
+++ b/hack/graph-util.py
@@ -421,13 +421,16 @@ def get_release_metadata(node):
         f.close()
 
         with tarfile.open(fileobj=io.BytesIO(layer_bytes), mode='r:gz') as tar:
-            f = tar.extractfile('release-manifests/release-metadata')
+            try:
+                f = tar.extractfile('release-manifests/release-metadata')
+            except KeyError:
+                continue
             meta = json.load(codecs.getreader('utf-8')(f))
             meta['image-config-data'] = image_config_data
             return meta
             # TODO: assert meta.get('kind') == 'cincinnati-metadata-v0'
 
-    raise ValueError('no release-metadata in {} layers ( {} )'.format(node['version'], json.dumps(manifest)))
+    raise ValueError('no release-metadata in {} layers ( {} )'.format(node['payload'], json.dumps(manifest)))
 
 
 if __name__ == '__main__':
--- a/hack/graph-util.py
+++ b/hack/graph-util.py
@@ -112,6 +112,7 @@ def load_nodes(directory, registry, repository):
             node['version'] = meta['version']
             if meta.get('previous'):
                 node['previous'] = set(meta['previous'])
+                node['internal-previous'] = set(meta['previous'])
             if meta.get('next'):
                 node['next'] = set(meta['next'])
             try:
@@ -264,8 +265,7 @@ def update_channels(node, token):
 def update_previous(node, token):
     labels = get_labels(node=node)
     if node.get('previous', set()):
-        meta = get_release_metadata(node=node)
-        previous = set(meta.get('previous', set()))
+        previous = node.get('internal-previous', set())
         want_removed = previous - node['previous']
         removed_label = labels.get('io.openshift.upgrades.graph.previous.remove', {})
         current_removed = set(version for version in removed_label.get('value', '').split(',') if version)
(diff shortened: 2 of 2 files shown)


wking:
has a cache-busting shift from `.nodes` to `.nodes-v2`, but should work.  I'll refine the cache busting to stay in `.nodes` if I have time, but doesn't seem like a big deal if folks want to land this with the rename approach.

wking:
huh:

```console
...
DEBUG: loaded from cache: 4.1.0-rc.0 quay.io/openshift-release-dev/ocp-release@sha256:345ec9351ecc1d78c16cf0853fe0ef2d9f48dd493da5fdffc18fa18f45707867
  File "hack/graph-util.py", line 408, in <module>
    push(directory='.', token=args.token, push_versions=args.versions)
  File "hack/graph-util.py", line 197, in push
    nodes = load_channels(directory=os.path.join(directory, 'channels'), nodes=nodes)
  File "hack/graph-util.py", line 135, in load_channels
    raise ValueError('{} claims version {}, but no nodes found with that version'.format(path, version))
ValueError: ./channels/stable-4.0.yaml claims version 4.0.0-0.9, but no nodes found with that version
```

Not sure what's up with that yet.  The release exists:

```console
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.0.0-0.9 | grep Name
Name:      4.0.0-0.9
```

But it's in my local cache:

```console
$ grep -r 4.0.0-0.9 .nodes-v2/
...no hits...
```

wking:
Ah, turns out 4.0.0-0.9 and friends have `schemaVersion` 1 manifests...

wking:
Ok, with f2f4a76, I think we're good to go here.  Merging with #14, this will push both the amd64 and s390x 4.2.14 into candidate-4.2.  If you wanted to, say, push only the amd64 release in, you'd use `4.2.14+amd64`, and that works too.  Still some room for improved logging, e.g.:

```
INFO: adding 4.2.14 to channels: candidate-4.2
INFO: 4.2.14 post https://quay.io/api/v1/repository/openshift-release-dev/ocp-release/manifest/sha256:26ba0c17618872c68a2e9a5002c3f3e48f75ae844fcf66e83063bd72a104cfc7/labels
```

doesn't tell you that's the s390x release, but it's late, and logging is not critical ;).

steveeJ:
This PR looks like it's duplicating more of what we already have implemented in the [Cincinnati graph-builder registry code](https://github.com/openshift/cincinnati/blob/master/graph-builder/src/registry.rs#L194). I really want to encourage us to start working on re-using that vs adding more functionality here.

wking:
> I really want to encourage us to start working on re-using that...

Sounds good to me, but 4.2.14 needs to get tagged into channels today, and I don't expect us to have a Rust port today.

README.md
```diff
@@ -14,6 +14,12 @@ The [contributing documentation](CONTRIBUTING.md) covers licencing and the usual
 
 **Do Not Ever Update Quay Labels Based On Your Local Development Branch. Only from master!**
```
LalatenduMohanty: Not a review comment but as a note, this will not be an issue when Cincinnati will only look in to git for graph data. It will also reduce the moving

hack/graph-util.py
```diff
@@ -60,12 +87,14 @@ def load_nodes(directory, registry, repository):
                     except ValueError as error:
                         raise ValueError('failed to load YAML from {}: {}'.format(path, error))
                     if not meta:
+                        if entry['name'] == '4.0.0-0.9':
```
eparis: ??

wking: whoops.  Dropping this debugging stuff...

wking: fixed with f2f4a76 -> beb55be.

README.md
```diff
@@ -14,6 +14,12 @@ The [contributing documentation](CONTRIBUTING.md) covers licencing and the usual
...
 
+### Release names
+
+When [adding releases to channels](#add-releases-to-channels) or [blocking edges](#block-edges), releases are given by name.
+Architecture-agnostic names like 4.2.14 will apply to all images with that exact name in the `version` property of the `release-metadata` file included in the release image.
+Names with [SemVer build metadata][semver-build] like 4.2.14+amd64 will apply only to releases whose exact name in the `version` property matches the release with the build metadata removed (e.g. 4.2.14) but whose referenced [image architecture][image-arch] matches the given build metadata.
```
LalatenduMohanty: The current arch specific releases have `4.2.12-s390x` as name, so we should use this an example instead of `4.2.14+amd64`?

LalatenduMohanty: Also `Names with [SemVer build metadata][semver-build] like 4.2.14+amd64 will apply only to releases whose exact name in the `version` property matches the release with the build metadata removed (e.g. 4.2.14) but whose referenced [image architecture][image-arch] matches the given build metadata.` is a long sentence and we should break this down.

wking: > The current arch specific releases have 4.2.12-s390x as name, so we should use this an example instead of 4.2.14+amd64?

`-` (prerelease name) instead of `+` (build metadata) was an artifact of old ART naming.  This commit has nothing to do with prerelease names, and folks handling those will still need to use the full name which was baked into the release (e.g. [this][1] is still correct).  With this PR, you could use `4.2.11-s390x+390x` as well, but since there are no releases named `4.2.11-s390x` on other arches, you don't have to.

[1]: https://github.com/openshift/cincinnati-graph-data/blob/8e681e27f8f725b8506839a39803ebc0e5aaf575/channels/stable-4.2.yaml#L20

LalatenduMohanty: Suggestions around breaking the sentences: 
* When adding releases to channels or blocking edges, releases should be given by name. 
* Architecture-agnostic names like 4.2.14 will apply to all images with that exact name in the version property of the release-metadata file included in the release image. 
* Names with SemVer build metadata like 4.2.14+amd64 will match to releases stings without the build metadata (e.g. 4.2.14)  in version property along with the referenced image architecture matches the given build metadata.

wking: >...is a long sentence and we should break this down.

I've pushed beb55be -> 3c49e24, pulling the 4.2.14 examples down into subsequent sentences.  How does that look?

eparis:
This is fine with me from a usage PoV. The only scary thing is if secondary arches ever decide to release on a different schedule. When pushed they would appear in the graph wherever the x86_64 release exists...

eparis:
Let me run that down with crawford/jessica to see if that is likely to happen...

README.md
```diff
@@ -14,6 +14,14 @@ The [contributing documentation](CONTRIBUTING.md) covers licencing and the usual
 
 **Do Not Ever Update Quay Labels Based On Your Local Development Branch. Only from master!**
 
+### Release names
+
+When [adding releases to channels](#add-releases-to-channels) or [blocking edges](#block-edges), releases are given by name.
```
LalatenduMohanty: s/releases are given by name/releases must be given by name/

wking: > s/releases are given by name/releases must be given by name/

This section is really about "when you put some characters in a release-name field, how are they interpreted", not about declaring "this is a release name field".  I've pushed 34cfee3 -> 572e1fe moving to:

```markdown
Release names are used for [adding releases to channels](#add-releases-to-channels) and [blocking edges](#block-edges).
```

to make this a bit more concise "here's why you care about this section".  How do you feel about that?

wking:
> The only scary thing is if secondary arches ever decide to release on a different schedule.

@sosiouxme has pointed out that divergent schedules gets tricky from an Errata/product-page perspective, but I'm not clear on the details.

PR #18, opened 2020-01-14
Add 4.2.14 to stable

joepvd:


channels/fast-4.2.yaml
```diff
@@ -25,3 +25,5 @@ versions:
 - 4.2.12-s390x
 - 4.2.13
 - 4.2.13-s390x
+- 4.2.14
+- 4.2.14-s390x
```
wking: We don't want s390x, because of [rhbz#1789260][1] (#17).  Can we make this (and the stable channel):

```yaml
- 4.2.14-amd64
# not 4.2.14-s390x because of https://bugzilla.redhat.com/show_bug.cgi?id=1789260
```

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1789260

wking:
Also, I'd personally rather add this to fast, wait a while to judge update success in the wild, and then promote to stable in a follow-up PR.  But that's really a release-admin call.

channels/fast-4.2.yaml
LalatenduMohanty: But was 4.2.14 is also supposed to be for s390x? It should not as we have the bug. Do we have the information around which arch we doing releases somewhere? I am worried about how can we avoid  adding wrong arch in these type of PRs.  The bug is not marked blocker too.

wking: > But was 4.2.14 is also supposed to be for s390x?

We should have excluded it from candidate-4.2, but did not because #14 and #16 kind of stumbled in without #14 being rebased around the parallel #17 landed.

wking:
oh, wait, we want to include the 4.1.z here too, right?  E.g. #14 put 4.1.30 in `candidate-4.2`, and now we want to also put it in `fast-4.2`.

PR #22, opened 2020-01-16
channels/stable: 4.1.30 and 4.2.14 (amd64) to stable

wking:
Not much changed since they went into fast, but we don't expect problems, so stabilize them (but not for s390x).


wking:
/hold

I want to take a closer look at 4.2.* -> 4.2.14.

eparis:
/lgtm
after trevor pulls the hold based on test

wking:
> after trevor pulls the hold based on test

There haven't been new 4.1.30 -> 4.2.14 updates since #21, so I don't see a point in waiting on that.  After reviewing Telemery from clusters that got stuck on 4.2.* -> 4.2.14, I'm personally comfortable with our understanding of the existing failure modes (e.g. one cluster hit [rhbz#1791934][1], but they can work around that by enabling the registry operator again or, since the fixed-in-4.2.10 [rhbz#1753778][2], setting it Removed).  The other failures seemed to be mostly that sort of recoverable-cluster-config thing.  There were ~three clusters that might have been "compute machine did not survive reboot", and while that's slightly concerning, there's not a lot of information in Telemetry/Insights about what went wrong, and compute failures should be easy to recover (if you have a MachineSet, you can remove the machine and the machine-API will make you a new one).

But I don't think I should be pulling the hold here.  If nothing else, we need a release admin to actually push this to Quay once it lands.  So I'll leave it to a release admin to remove the hold when they're comfortable.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1791934
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1753778

wking:
Rebased onto master (I'd accidentally put it on top of #21 before).  That clears the LGTM, so I can:

/hold cancel

and still not have this merge without new release-admin approval.

PR #25, opened 2020-01-16
Read Quay token from file as well as from command line

eparis:


hack/graph-util.py
```diff
@@ -346,7 +346,6 @@ def delete_label(node, label, token, key=None):
     request.get_method = lambda: 'DELETE'
     return urlopen(request)
 
-
```
wking: nit: [PEP 8][1]:

> Surround top-level function and class definitions with two blank lines.

Can we keep this one?

[1]: https://www.python.org/dev/peps/pep-0008/#blank-lines

hack/graph-util.py
```diff
@@ -467,5 +477,6 @@ def get_release_metadata(node):
 
     args = parser.parse_args()
 
+    token = get_token(args)
```
wking: nit (and this is super tiny): I like keyword argument like `args=args` to make it harder to accidentally pass arguments in the wrong order.  In the other hand, it seems very unlikely that `get_tokens` ever grows more positional arguments.

wking:
/lgtm
/hold

Two minor nits that you can fix or not as you see fit.  Pull the hold if you want to ignore them, and ping me if you fix them and I'll re-LGTM

PR #26, opened 2020-01-16
candidate-4.3: add 4.3.0-rc.2

sosiouxme:


wking:
I think rc.2 was fine, but no need to add it now that rc.3 has leapfrogged it into candidate-4.3 #28 

/close

PR #23, opened 2020-01-16
channels/{prerelease,candidate}: Add 4.1.31 and 4.2.15

joepvd:

PR #29, opened 2020-01-21
candidate-4.2: add 4.2.16 and 4.1.31

sosiouxme:


wking:
Replacing #23.  Probably should have had a comment in here about why we did not promote 4.2.15.

PR #30, opened 2020-01-22
channels: Create stable-4.3 with 4.3.0-x86_64

thiagoalessio:
Related ticket: https://issues.redhat.com/browse/OCPPLAN-2630

### DIFF
--- a/channels/stable-4.3.yaml
+++ b/channels/stable-4.3.yaml
@@ -1,0 +1,3 @@
+name: stable-4.3
+versions:
+- 4.3.0-x86_64


eparis:
This would need to go to candidate-4.3, not stable. the auto updates team, I think, takes the lead getting stuff from candidate to fast/stable.  @LalatenduMohanty @wking is this true?

Also, if there is no s390/powerpc build you can just do 4.3.0, no need for the `-x86_64`.  If there is an s390/powerpc build, which we are trying to intentionally exclude, the format would be `4.3.0+amd64`

LalatenduMohanty:
> This would need to go to candidate-4.3, not stable. the auto updates team, I think, takes the lead getting stuff from candidate to fast/stable. @LalatenduMohanty @wking is this true?

It should first go to the candidates channel then `fast channel` then `stable`

sosiouxme:
Clearly ART docs need updates :)
https://github.com/openshift/cincinnati-graph-data/pull/31 to put it in candidate first.

PR #31, opened 2020-01-22
candidate-4.3: add 4.3.0

sosiouxme:


LalatenduMohanty:
@eparis +1 to merge this. I will not `lgtm` this as IMO only release admins should merge this till we auto merge this in future based on some criteria.

wking:
This should probably include "we're not adding 4.2.16 to candidate-4.3 because of the GCP scheduler issues: https://bugzilla.redhat.com/show_bug.cgi?id=1785457" in a YAML comment.

wking:
And probably also include an explicit blocked edge file

PR #13, opened 2020-01-08
channels/candidate-4.3: Promote 4.3.0-rc.0-x86_64

wking:
New channel for the new 4.3.z-stream.

PR #32, opened 2020-01-22
Blocking all the edges to 4.3.0 because of bz 1793635

LalatenduMohanty:
https://bugzilla.redhat.com/show_bug.cgi?id=1793635


wking:
Probably for rc.3 as well, right?  With at least the GCP bug, and also potentially any of [these][1] that were unfixed at that point (or a comment about them being fixed by rc.3?).

[1]: https://github.com/openshift/cincinnati-graph-data/pull/13#issuecomment-573939250

LalatenduMohanty:
> Probably for rc.3 as well, right? With at least the GCP bug, and also potentially any of [these](https://github.com/openshift/cincinnati-graph-data/pull/13#issuecomment-573939250) that were unfixed at that point (or a comment about them being fixed by rc.3?).

That would be a different PR.

PR #38, opened 2020-01-28
stable-4.2: add 4.2.16

sosiouxme:
also 4.1.31 for incoming edges


wking:
Oh, wait, these should be promoting 4.1.31 into stable-4.2 too (and fast-4.2, to catch up with #37).  Or we can handle that in a separate PR, but without it folks could be sitting on the 4.1 tip without a stable path over to 4.2.

PR #40, opened 2020-02-03
Enable the 4.2.16 to 4.3.0 upgrade

jwforres:


eparis:
4.2.16 is already in fast-4.2, so this LGTM.
/lgtm
/hold in case we care about timing of the merge.

PR #41, opened 2020-02-03
Do not add 4.2 s390 release to 4.3 channels

eparis:


eparis:
I added comments to the files. So I manually re-lgtm'd it.

PR #42, opened 2020-02-04
Print the architecture in log messages

eparis:
We have introduced architectures into channels by mistake because the logs
talk about the 'release' but don't make it obvious that release might be for
multiple archs. This prints the versions similar to how they would have to be
listed in the config files if you wanted arch specificity.

### DIFF
--- a/hack/graph-util.py
+++ b/hack/graph-util.py
@@ -30,6 +30,16 @@ logging.basicConfig(format='%(levelname)s: %(message)s')
 _LOGGER = logging.getLogger(__name__)
 _LOGGER.setLevel(logging.DEBUG)
 
+def get_architecture(meta):
+    return meta['image-config-data']['architecture']
+
+def get_log_version(node):
+    meta = node['meta']
+    arch = get_architecture(meta)
+    semver_match = _VERSION_REGEXP.match(node['version'])
+    if not semver_match:
+        raise ValueError('version {} is not a valid Semantic Version'.format(node['version']))
+    return '{}+{}'.format(remove_build(semver_match=semver_match), arch)
 
 def remove_build(semver_match):
     data = semver_match.groupdict()
@@ -88,7 +98,8 @@ def load_nodes(directory, registry, repository):
                         raise ValueError('failed to load YAML from {}: {}'.format(path, error))
                     if not meta:
                         continue
-                _LOGGER.debug('loaded from cache: {} {}'.format(meta['version'], node['payload']))
+                arch = get_architecture(meta)
+                _LOGGER.debug('loaded from cache: {}+{} {}'.format(meta['version'], arch, node['payload']))
             except IOError:
                 try:
                     meta = get_release_metadata(node=node)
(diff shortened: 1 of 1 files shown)


hack/graph-util.py
```diff
@@ -30,6 +30,10 @@
 _LOGGER = logging.getLogger(__name__)
 _LOGGER.setLevel(logging.DEBUG)
 
+def get_log_version(node):
+    if node['meta']['image-config-data']['architecture']:
```
vrutkovs:
```suggestion
    if node['meta'].get('image-config-data', {}).get('architecture'):
```
so that it would work with images where `image-config-data` is not set

vrutkovs: Since `meta['image-config-data']['architecture']` is used in other places perhaps a better solution would be creating `def get_architecture(meta)` instead?

eparis:
more readable like this?

hack/graph-util.py
```diff
@@ -30,6 +30,13 @@
 _LOGGER = logging.getLogger(__name__)
 _LOGGER.setLevel(logging.DEBUG)
 
+def get_architecture(meta):
+    return meta['image-config-data']['architecture']
```
vrutkovs:
```suggestion
    return meta.get('image-config-data', {}).get('architecture')
```
In case `meta` doesn't have these

vrutkovs: Although I'm not sure - do we still have releases without arch info? @wking or @steveeJ ?

wking: > ...do we still have releases without arch info...

Arch info is extracted from the container image itself; there's no way you can avoid having it.  Sanity check:

```console
DEBUG: caching metadata for 4.2.18 quay.io/openshift-release-dev/ocp-release@sha256:144485daf08522e7fab77cb90ce360e72aca7a92f633a74b522808b4fe2b6627
...
$ grep -chr architecture .nodes | sort | uniq -c
      1 0
    100 1
$ grep -cr architecture .nodes | grep ':0$'
.nodes/version:0
```

(i.e. "the only file under `.nodes` without an `architecture` is the cache-version file").  So I'm fine with this as @eparis has it in 0f777c88b.

wking: Although personally I would not have bothered wrapping this one-liner in a `get_architecture` function, but whatever ;)

hack/graph-util.py
```diff
@@ -30,6 +30,13 @@
...
+
+def get_log_version(node):
+    meta = node['meta']
+    arch = get_architecture(meta)
+    return "%s+%s" % (node['version'], arch)
```
wking: You probably want:

```python
semver_match = _VERSION_REGEXP.match(node['version'])
if not semver_match:
    raise ValueError('version {} is not a valid Semantic Version'.format(node['version']))
return '{}+{}'.format(remove_build(semver_match=semver_match), arch)
```

to guard against doubling the build metadata.  Although we have not actually cut a release with build metadata (yet).  I'm also fine with rendering this as `{} ({}).format(node['version'], arch)`, and not trying to cram it into SemVer.

hack/graph-util.py
wking: hah, and I see @vrutkovs explicitly asking for a `get_architecture` wrapper [here][1].

[1]: https://github.com/openshift/cincinnati-graph-data/pull/42#discussion_r374623634

vrutkovs: Ah, great, lets keep it like this then

hack/graph-util.py
wking: 0f777c8 is using a `get_architecture` wrapper.

hack/graph-util.py
eparis: I don't care how we do it, just so long as when we look at the output we know what arch's changes are being applied to.  So I just took your suggestion.

PR #45, opened 2020-02-06
channels/candidate-4.3: Promote 4.2.16+amd64

wking:
792cc4c3d7 (#40) pulled the block to allow this edge, but to get an edge in a given channel, both the source and target release need to be in that channel.

See also #44, which goes into more detail on 4.2->4.3 failure modes, some of which still occur on 4.2.16->4.3.0.

### DIFF
--- a/channels/candidate-4.3.yaml
+++ b/channels/candidate-4.3.yaml
@@ -1,5 +1,7 @@
 name: candidate-4.3
 versions:
+# until s390 is released on 4.3 we may not want to include it in 4.3 channels
+- 4.2.16+amd64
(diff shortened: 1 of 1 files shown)


channels/candidate-4.3.yaml
```diff
@@ -1,5 +1,6 @@
 name: candidate-4.3
 versions:
+- 4.2.16
```
eparis: in fast-4.3 we did 4.2.16+amd64

wking: Fixed with 0a03965, which also links #41.

PR #47, opened 2020-02-07
WIP: hack/graph-util: Add 'print' to write JSON to stdout

wking:
To allow:

```console
$ hack/graph-util.py print >graph.json
```

to dump our assembled metadata to disk, as an alternative to passing it to Cincinnati via Quay labels.

### DIFF
--- a/hack/graph-util.py
+++ b/hack/graph-util.py
@@ -10,6 +10,7 @@ import multiprocessing.dummy
 import os
 import re
 import shutil
+import sys
 import tarfile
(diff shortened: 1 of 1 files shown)


eparis:
/lgtm
How hard would a printer be that gave Cincinnati compatible outputs?

wking:
/close

Stefan is going to grow a graph-builder plugin to parse this Git repo directly: openshift/cincinnati#231.

PR #48, opened 2020-02-09
Return a better error if you forget to add the subcommand

eparis:
before it would just throw and exception. Now it complains that no command
was found and says the only valid one is {'push-to-quay'}

### DIFF
--- a/hack/graph-util.py
+++ b/hack/graph-util.py
@@ -471,7 +471,7 @@ def get_token(args):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Utilities for managing graph data.')
-    subparsers = parser.add_subparsers()
+    subparsers = parser.add_subparsers(dest='command')
 
     push_to_quay_parser = subparsers.add_parser(
         'push-to-quay',
@@ -489,10 +489,11 @@ if __name__ == '__main__':
         '--versions',
         help='Comma Seperated Versions to sync',
     )
-    push_to_quay_parser.set_defaults(action='push-to-quay')
 
     args = parser.parse_args()
 
-    token = get_token(args=args)
-    if args.action == 'push-to-quay':
+    if args.command == 'push-to-quay':
+        token = get_token(args=args)
         push(directory='.', token=token, push_versions=args.versions)
+    else:
+        parser.print_usage()


hack/graph-util.py
```diff
@@ -456,7 +456,7 @@ def get_token(args):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Utilities for managing graph data.')
-    subparsers = parser.add_subparsers()
+    subparsers = parser.add_subparsers(dest='command', required=True)
```
wking: `required` is [new in 3.7][1].  How about:

```diff
@@ -456,7 +456,7 @@ def get_token(args):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Utilities for managing graph data.')
-    subparsers = parser.add_subparsers()
+    subparsers = parser.add_subparsers(dest='command')
 
     push_to_quay_parser = subparsers.add_parser(
         'push-to-quay',
@@ -474,10 +474,11 @@ if __name__ == '__main__':
         '--versions',
         help='Comma Seperated Versions to sync',
     )
-    push_to_quay_parser.set_defaults(action='push-to-quay')
 
     args = parser.parse_args()
 
-    token = get_token(args=args)
-    if args.action == 'push-to-quay':
+    if args.command == 'push-to-quay':
+        token = get_token(args=args)
         push(directory='.', token=token, push_versions=args.versions)
+    else:
+        parser.print_usage()
```

or some such?  With my Python 2.7.5 and 3.4.9, that gives:

```console
$ python2 hack/graph-util.py 
usage: graph-util.py [-h] {push-to-quay} ...
graph-util.py: error: too few arguments
$ python3 hack/graph-util.py 
usage: graph-util.py [-h] {push-to-quay} ...
```

Not clear to me why my Python 2 is pickier (dying in `parse_args` instead of getting through to the patched `print_usage`).  But as long as it's obvious that no pushing to Quay or anything happened, I don't think that distinction matters.

[1]: https://docs.python.org/3/library/argparse.html#sub-commands

eparis: Darn you python 2. Ok, I'll rework it. Thx Trevor.

eparis: fixed.

PR #49, opened 2020-02-12
candidate-4.2: add 4.2.18

thiagoalessio:


channels/candidate-4.3.yaml
```diff
@@ -2,6 +2,7 @@ name: candidate-4.3
 versions:
 # until s390 is released on 4.3 we may not want to include it in 4.3 channels
 - 4.2.16+amd64
+- 4.2.18
```
eparis: Do we want +amd64?

thiagoalessio: oh yes, sorry ... added!

PR #59, opened 2020-02-19
blocked-edges/4.3.0: Block 4.2.16 and earlier updating to 4.3

wking:
The 4.2 machine-config fix was [rhbz#1782152][1], [landed in bd358bb][2], which is new in 4.2.18:

```console
$ oc adm release info --commits quay.io/openshift-release-dev/ocp-release:4.2.16-x86_64 | grep machine-config
  machine-config-operator                       https://github.com/openshift/machine-config-operator                       31fed93186c9f84708f5cdfd0227ffe4f79b31cd
$ oc adm release info --commits quay.io/openshift-release-dev/ocp-release:4.2.18-x86_64 | grep machine-config
  machine-config-operator                       https://github.com/openshift/machine-config-operator                       9366460085b2a24d825380759f554769ec5ab4f9
$ git --no-pager log --oneline --first-parent -2 9366460085
93664600 Merge pull request #1362 from rphillips/fixes/1787581_4.2
bd358bb7 Merge pull request #1323 from openshift-cherrypick-robot/cherry-pick-1320-to-release-4.2
```

There was no 4.2.17; @vikaslaad on 2020-02-10:

> 4.2.17 was skipped due to multiple weeks in between

I'm not entirely clear on the motivation there, but have added a comment about the skip with an attempt at explaining it.

Updates also need the 4.3 fix from [rhbz#1782149][3], [landed in 9fd53bd][4], which landed early enough for 4.3.0-rc.0:

```console
$ oc adm release info --commits quay.io/openshift-release-dev/ocp-release:4.3.0-rc.0-x86_64 | grep machine-config
  machine-config-operator                       https://github.com/openshift/machine-config-operator                       23a6e6fb37e73501bc3216183ef5e6ebb15efc7a
$ git --no-pager log --oneline --first-parent -8 23a6e6fb37
23a6e6fb Merge pull request #1348 from openshift-cherrypick-robot/cherry-pick-1285-to-release-4.3
80c8aed7 Merge pull request #1343 from retroflexer/cherry-pick-backup-restore-kube-static-resources
269990a3 Merge pull request #1344 from openshift-cherrypick-robot/cherry-pick-1296-to-release-4.3
fd3ca395 Merge pull request #1338 from runcom/fix-go-mod
ba304dbb Merge pull request #1333 from openshift-cherrypick-robot/cherry-pick-1278-to-release-4.3
787f3fa9 Merge pull request #1332 from runcom/reserved-cpus-4.3
[...]

PR #43, opened 2020-02-05
candidate-4.3: add 4.3.1

sosiouxme:

PR #60, opened 2020-02-19
channels/fast-4.2: Promote 4.2.18 (and 4.2.18+amd64 to fast-4.3)

wking:
The release has been in the candidate analogs since 6ed3a35fa1 (#49), which is plenty of cook time.


wking:
Rebased to sit on top of #61 instead, because #59 is turning out to be complicated.  We can add 4.2.18 to fast while we work out what to do with 4.2.16.

wking:
Also revert #43, now that we actually have a 4.2.18 release.

PR #62, opened 2020-02-20
channels: Explain why we skipped 4.1.32 and 4.1.33

wking:
I actually have no idea why we skipped these, but 4.1.34 is the next release after 4.1.31.


wking:
@thiagoalessio:

> starting with 4.1.31
>
> > Note: 4.1 releases will no longer be weekly, but now "as needed" for critical fixes.
>
> but we still have our Friday automation turned on, so it will keep bumping the numbers every week, releasing it or not.

PR #65, opened 2020-02-21
channels/stable-4.2: 4.2.18 to stable

thiagoalessio:


wking:
Does not include stable-4.3, so we don't need to worry about 4.2.18 -> 4.3.  No later 4.2 in stable-4.2, so don't need to worry about 4.2.18 -> 4.2.  Checking baked in update sources:

```console
$ for ARCH in x86_64 s390x; do printf "${ARCH}\t"; oc adm release info "quay.io/openshift-release-dev/ocp-release:4.2.18-${ARCH}" | grep Upgrades; done
x86_64      Upgrades: 4.1.34, 4.2.7, 4.2.8, 4.2.9, 4.2.10, 4.2.12, 4.2.13, 4.2.14, 4.2.15, 4.2.16
s390x      Upgrades: 4.1.34, 4.2.7, 4.2.8, 4.2.9, 4.2.10, 4.2.12, 4.2.13, 4.2.14, 4.2.15, 4.2.16
```

4.1.34 isn't in stable-4.2 either, so the only edges we need to look at are 4.2 -> 4.2.18.  Looking at telemetry for 4.2 -> 4.2.18, things seem pretty quiet.  But nothing is jumping out at me as a serious issue yet, so I think we can safely add to stable-4.2 and pull the edge later if something comes up as traffic trickles over.

/lgtm

wking:
No hold for 6h, and PR is still open?

/test publish

PR #64, opened 2020-02-21
channels/fast-4.2: Promote 4.2.19 (and 4.2.19+amd64 to fast-4.3)

thiagoalessio:

PR #66, opened 2020-02-21
channels/stable-4.2: 4.2.19 to stable

thiagoalessio:


wking:
/hold

Blocked on fast in #64

eparis:
does it belong in stable-4.3?

vrutkovs:
Blocked by https://github.com/openshift/cincinnati-graph-data/pull/64 (`fast-4.2`)

eparis:
/approve cancel
I think this should go into stable-4.3 at the same time

LalatenduMohanty:
#64 is merged now. @thiagoalessio Can you rebase this PR?

LalatenduMohanty:
@eparis It is in https://github.com/openshift/cincinnati-graph-data/blob/master/channels/fast-4.3.yaml#L8 but not in stable-4.3 as we do not have any 4.2 release in stable-4.3

eparis:
Why are there no 4.2 releases in 4.3-stable?

LalatenduMohanty:
> Why are there no 4.2 releases in 4.3-stable?

We have not enabled edges in stable because of https://github.com/openshift/cincinnati-graph-data/blob/master/blocked-edges/4.3.0.yaml#L4

PR #67, opened 2020-02-21
channels/fast-4.2: Promote 4.2.20 (and 4.2.20+amd64 to fast-4.3)

thiagoalessio:

PR #68, opened 2020-02-21
channels/stable-4.2: 4.2.20 to stable

thiagoalessio:


vrutkovs:
/hold

Depends on https://github.com/openshift/cincinnati-graph-data/pull/67

wking:
/hold

Blocked on fast in #67

eparis:
/approve
/hold
but should it be in stable-4.3 as well?

eparis:
/approve cancel
I think this should go into stable-4.3 at the same time (but only the amd64 version)

LalatenduMohanty:
> I think this should go into stable-4.3 at the same time (but only the amd64 version)
The plan was to introduce 4.2 builds in stable-4.3 after 4.3.3 but now has blocker issues we have to wait for 4.3.5

PR #74, opened 2020-02-24
raw/metadata.json: add historical metadata from quay

steveeJ:
The metadata in this has been manually added on quay in the past.  It's
my understanding that we don't plan to a convenient mapping for adding
edges using `previous.add`, so we don't want a convenience structure for
this.

---
 
This is part of https://github.com/openshift/cincinnati/pull/226.

### DIFF
--- a/raw/OWNERS
+++ b/raw/OWNERS
@@ -1,0 +1,10 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+# This file just uses aliases defined in OWNERS_ALIASES.
+
+approvers:
+  - release-admins
+reviewers:
+  - cincinnati-graph-data-approvers
+  - cincinnati-graph-data-reviewers
+options:
+  no_parent_owners: true
--- a/raw/metadata.json
+++ b/raw/metadata.json
@@ -1,0 +1,14 @@
+{
+    "4.1.0": {
+      "io.openshift.upgrades.graph.previous.add": "4.1.0-rc.4,4.1.0-rc.9"
+    },
+    "4.1.0-rc.6": {
+      "io.openshift.upgrades.graph.previous.add": "4.1.0-rc.5"
+    },
+    "4.1.0-rc.7": {
+      "io.openshift.upgrades.graph.previous.add": "4.1.0-rc.6"
+    },
+    "4.1.0-rc.9": {
+      "io.openshift.upgrades.graph.previous.add": "4.1.0-rc.7,4.1.0-rc.8"
+    }
+}


raw_metadata.json
```diff
@@ -0,0 +1,20 @@
...
+        "io.openshift.upgrades.graph.previous.add": "4.1.0-rc.7"
+    },
+    "4.1.0-rc.9+amd64": {
+        "io.openshift.upgrades.graph.previous.add": "4.1.0-rc.7,4.1.0-rc.8"
+    },
+    "4.1.0+amd64": {
+        "io.openshift.upgrades.graph.previous.add": "4.1.0-rc.4,4.1.0-rc.9"
+    },
+    "4.2.13-s390x+s390x": {
+        "io.openshift.upgrades.graph.previous.add": "4.2.10-s390x,4.2.12-s390x"
```
wking: We can drop this from this PR and from Quay labels, because 4.2.13-s390x has [blocked all incoming edges][1].

[1]: https://github.com/openshift/cincinnati-graph-data/blob/04c5b057d2fbc8d235d86a32b46cb4dda891023c/blocked-edges/4.2.13-s390x.yaml#L3

wking: @jwforres [removed the label][1], so we can drop this here.

[1]: https://github.com/wking/cincinnati-graph-data/commit/07ffd741320d25026340e5f43367e4f56ec0e177

raw_metadata.json
```diff
@@ -0,0 +1,17 @@
+{
+    "4.1.0-rc.7+amd64": {
+        "io.openshift.upgrades.graph.previous.add": "4.1.0-rc.6"
+    },
+    "4.1.0-rc.6+amd64": {
+        "io.openshift.upgrades.graph.previous.add": "4.1.0-rc.5"
+    },
+    "4.1.0-rc.8+amd64": {
+        "io.openshift.upgrades.graph.previous.add": "4.1.0-rc.7"
```
wking: This is redundant, because the 4.1.0-rc.7 -> 4.1.0-rc.8 edge is already compiled into the 4.1.0-rc.8 release.

```console
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.1.0-rc.8 | grep Upgrades
  Upgrades: 4.1.0-rc.7
```

We can drop the label from Quay without affecting the graph.

wking: @jwforres [removed this one too][1].

[1]: https://github.com/wking/cincinnati-graph-data/commit/ddc929278ec36bb2b2ad434db99994e7f292fbdb

wking:
For context, we need these post-release edge injections to preserve this portion of the `prerelease-4.1` Cincinnati graph:

Screenshot from 2020-02-24 11-13-55

because those edges are not included in the release image metadata (which is our current best-practice):

```console
$ for TAG in 4.1.0-rc.6 4.1.0-rc.7 4.1.0-rc.9 4.1.0; do echo -ne "${TAG}\t"; oc adm release info "quay.io/openshift-release-dev/ocp-release:${TAG}" | grep Upgrades; done
4.1.0-rc.6	  Upgrades: <none>
4.1.0-rc.7	  Upgrades: <none>
4.1.0-rc.9	  Upgrades: <none>
4.1.0	  Upgrades: <none>
```

/lgtm

(but will need a release admin to approve).

wking:
> ...(but will need a release admin to approve).

or not, because @steveeJ was authorized to approve the change being made by the initial PR?  Or because he's allowed to create new directories that get delegated to other approvers?  Or something like that.

PR #75, opened 2020-02-25
channels: Add 'feeder' and 'tombstones' properties

wking:
So we can automatically propose stabilizing channel promotions.

The implementation uses `git blame --first-parent ...` to determine when releases landed in the feeder channel.  Picking the versions out of the blame data by using a '- ' prefix is a bit of a hack, but it gets the job done for now.


wking:
Example output (as of 0c41d5791f41):

```console
$ hack/graph-util.py stabilization-changes
INFO: considering promotions from candidate-4.2 to fast-4.2 after PT24H
INFO:   recommended: 4.1.18 (81 days, 1:33:57.131135)
DEBUG:     channels/fast-4: Promote 4.1.18 to fast-4.2
DEBUG:     It was promoted the feeder candidate-4.2 by 2c088bc147 (channels: Convert to Eric's proposed channel format, 2019-12-05).
INFO:   recommended: 4.1.20 (81 days, 0:53:55.131135)
DEBUG:     channels/fast-4: Promote 4.1.20 to fast-4.2
DEBUG:     It was promoted the feeder candidate-4.2 by b4db5a619e (channels: Catch up with production, 2019-12-05).
...
```

wking:
Example output (as of dbbbbd7a8fb):

```console
$ hack/graph-util.py stabilization-changes 2>&1 
INFO: considering promotions from candidate-4.2 to fast-4.2 after PT24H
INFO:   recommended: 4.1.34 (12 days, 13:34:53.517678)
DEBUG:     channels/fast-4.2: Promote 4.1.34 to fast-4.2
DEBUG:     It was promoted the feeder candidate-4.2 by 8e0a12f267 (candidate-4.1: add 4.1.34, 2020-02-12).
INFO:   recommended: 4.2.20 (5 days, 13:04:12.517678)
DEBUG:     channels/fast-4.2: Promote 4.2.20 to fast-4.2
DEBUG:     It was promoted the feeder candidate-4.2 by 3eb8b2e776 (candidate-4.2: add 4.2.20, 2020-02-19).
...
```

README.md
```diff
@@ -25,13 +25,49 @@ And 4.2.14+amd64 would only apply to the amd64 release image.
 ### Add Releases To Channels
 
 Edit the appropriate file in `channels/`.
-For example, to add a release to stable-4.2 you would edit `channels/stable-4.2.yaml`.
-Channel semantics are documented [here][channel-semantics].
+For example, to add a release to candidate-4.2 you would edit [`channels/candidate-4.2.yaml`](channels/candidate-4.2.yaml).
```
LalatenduMohanty: s/would/need to/

README.md
```diff
@@ -25,13 +25,49 @@ And 4.2.14+amd64 would only apply to the amd64 release image.
...
 The file contains a list of versions.
 Please keep the versions in order.
 And LEAVE COMMENTS if you skip a version.
 
+#### Feeder Channels
+
+Channel semantics, as documented [here][channel-semantics], show nodes and edges being promoted to successive channels as they prove their stability.
+For example, a 4.2.z release will appear in `candidate-4.2` first.
+Upon proving itself sufficiently stable in the candidate channel, it will be promoted into `fast-4.2`.
+Some time after landing in `fast-4.2`, it will appear in `stable-4.2`.
```
LalatenduMohanty: Text suggestion:
Upon proving itself sufficiently stable in the candidate channel, it will be promoted into `fast-4.2`.
After the release stays in fast channel, it will appear in `stable-4.2`.
*Note*:  We are doing the delayed publish of builds in to stable till we have phased rollout in stable channel. Once we have the phased rollout capability we intend to publish a release to both fast and candidate release at one go.

README.md
```diff
@@ -25,13 +25,49 @@ And 4.2.14+amd64 would only apply to the amd64 release image.
...
+
+```yaml
+feeder:
+  name: candidate-4.2
+  delay: PT24H
+```
+
+which declares the intention that nodes and edges will be considered for promotion into `fast-4.2` after cooking for 24 hours in `candidate-4.2`.
+The `delay` value is an [ISO 8601][rfc-3339-p13] [duration][iso-8601-durations].
+
```
LalatenduMohanty: :+1:

README.md
wking: preexisting language; file a separate PR?

README.md
wking: Not quite your wording, but added a note about phased release rollouts in 9b9497b7b448.  File another comment if you have issues with the new wording.

README.md
```diff
@@ -25,13 +25,52 @@ And 4.2.14+amd64 would only apply to the amd64 release image.
...
+This is the expected delay, but it does not mean that promotion will happen at that moment.
```
steveeJ: By which automation is this delay used and how would one override it?

README.md
```diff
@@ -25,13 +25,52 @@ And 4.2.14+amd64 would only apply to the amd64 release image.
...
+
+Channel semantics, as documented [here][channel-semantics], show nodes and edges being promoted to successive channels as they prove their stability.
+For example, a 4.2.z release will appear in `candidate-4.2` first.
+Upon proving itself sufficiently stable in the candidate channel, it will be promoted into `fast-4.2`.
+Some time after landing in `fast-4.2`, it will appear in `stable-4.2`.
+
+_Note:_ Once we have phased release rollouts, we will drop the fast/stable distinction from this repository and promote to a unified fast/stable channel with a start time and rollout duration.
+Until then, we are using fast channels to feed stable channels with a delay, just like candidate channels feed fast channels.
+
+In this repository, the intended promotion flow is reflected by a `feeder` property in the channel declaration.
```
steveeJ: Do we still need the `feeder` property if we reduce the chain to being _candidate_ --> _stable_?

README.md
```diff
@@ -25,13 +25,52 @@ And 4.2.14+amd64 would only apply to the amd64 release image.
...
+
+##### Tombstones
+
+Removing a node from a channel can strand existing clusters with a `VersionNotFound` error.
+To avoid that, unstable nodes are left in their existing channels, but should not be promoted to additional channels.
+This is reflected through entries in the optional `tombstones` property.
+For example, [`channels/candidate-4.2.yaml`](channels/candidate-4.2.yaml) has:
+
+```yaml
+tombstones:
```
steveeJ: I propose naming it _frozen_ instead of _tombstones_.

LalatenduMohanty: @steveeJ `frozen` does not communicate its intent. These are the nodes which can not have any future edge. So `tombstones` makes more sense.

LalatenduMohanty:
@eparis do we have comments on this PR.  I am planning to merge this soon.

README.md
steveeJ: I think that _tombstones_ doesn't communicate it well either, since these versions are still in use somewhere, which is the reason why we need to give them an additional property. Also, It would be nice to stick to a more technical or neutral term here.

How about _immobilized_?

LalatenduMohanty:
> @eparis do we have comments on this PR. I am planning to merge this soon.

There is a change of plan as discussed with OTA team.  We are planning to deprecate the python script so not planning to add new functionality to the script.

steveeJ:
> We are planning to deprecate the python script so not planning to add new functionality to the script.

I second this. I see the scope of this PR to discuss the data model changes and their intended usage, therefore please see my review comments.

README.md
vrutkovs: I think its still useful, just in case the names would be less clear

wking:
Rebased onto master with 9b9497b -> b3b6d48.

README.md
vrutkovs: I like tombstones more. These builds are similar to ghosts we don't speak of since they are dead to us and we don't want them coming back :)

README.md
wking: > By which automation is this delay used...

By a periodic that runs `stabilization-changes` and creates PRs for any recommended promotions.  Then eventually, when we trust our automatic edge-pulling, we give the robot creds to approve their own PRs so they will auto-merge.

> ... and how would one override it?

Manually create pull requests if you don't want to wait the configured delay?

README.md
wking: > Do we still need the feeder property if we reduce the chain to being candidate --> stable?

We also have prerelease -> stable for 4.1.  And we could theoretically also make a plural `feeders` and have multiple `candidate-4.*` feeding into `candidate` (#112).  I like this being explicit, but don't really care if folks would rather bake the feeder directions into the recommending script (currently the `stabilization-changes` subcommand) instead of tracking it in the channel YAML.

README.md
wking: Tombstone has [an existing technical usage][1], which is similar to the usage here.

[1]: https://en.wikipedia.org/wiki/Tombstone_(data_store)

wking:
>  We are planning to deprecate the python script so not planning to add new functionality to the script.

We are planning to deprecate the push-to-quay logic, which currently lives in the Python script.  This promotion-recommendation logic is distinct; it's about when to create pull requests for this repository.  I think we want it, and I don't think we want to include it in Cincinnati.  I don't care if folks want to rewrite it in Rust or not.  I can also put it in a separate Python script, instead of having it live in a subcommand of the existing script, if that would make folks more comfortable.  Thoughts?

wking:
Rebased again with ab9d6aa -> 15c6d24.

wking:
> I can also put it in a separate Python script...

Done with 7470db7 -> c69dc22.

channels/candidate-4.2.yaml
```diff
@@ -52,3 +52,11 @@ versions:
 - 4.2.21
 - 4.2.22
 - 4.2.23
+tombstones:
```
jwforres: Is there a reason 4.2.11 and 4.2.11-s390x were not included in the tombstones? We wouldn't want those promoted either right?

jwforres: Also I'd expect a reason / BZ for any release we choose to tombstone

channels/fast-4.3.yaml
```diff
@@ -1,4 +1,7 @@
 name: fast-4.3
+feeder:
+  name: candidate-4.3
```
jwforres: how are we handling n-1 minor feeders? both candidate-4.2 and candidate-4.3 feed into fast-4.3 (after we open the minor version upgrades for that release)

README.md
jwforres: If / when robots were auto-merging these PRs, how would release admins override the bot from promoting something

channels/candidate-4.2.yaml
wking: > Is there a reason 4.2.11 and 4.2.11-s390x were not included in the tombstones? We wouldn't want those promoted either right?

We had already promoted those before we noticed the problem:

```console
$ grep -A1 '4\.2\.11' channels/*-4.2.yaml 
channels/candidate-4.2.yaml:# - 4.2.11 failed to run tests, we never officially released it, but we accidentally
channels/candidate-4.2.yaml-# put it in a channel! (same for s390x)   Now we shouldn't pull it, just in case
--
channels/candidate-4.2.yaml:- 4.2.11
channels/candidate-4.2.yaml:- 4.2.11-s390x
channels/candidate-4.2.yaml-- 4.2.12
--
channels/fast-4.2.yaml:# - 4.2.11 failed to run tests, we never officially released it, but we accidentally
channels/fast-4.2.yaml-# put it in a channel! (same for s390x)   Now we shouldn't pull it, just in case
--
channels/fast-4.2.yaml:- 4.2.11
channels/fast-4.2.yaml:- 4.2.11-s390x
channels/fast-4.2.yaml-- 4.2.12
--
channels/stable-4.2.yaml:# - 4.2.11 failed to run tests, we never officially released it, but we accidentally
channels/stable-4.2.yaml-# put it in a channel! (same for s390x)   Now we shouldn't pull it, just in case
--
channels/stable-4.2.yaml:- 4.2.11
channels/stable-4.2.yaml:- 4.2.11-s390x
channels/stable-4.2.yaml-- 4.2.12
```

> Also I'd expect a reason / BZ for any release we choose to tombstone

I've been adding that for new tombstones (e.g. c69dc22d241), but hadn't gone back through internal discussion to see why other releases were tombstoned.  Should I do that digging?

channels/fast-4.3.yaml
wking: > how are we handling n-1 minor feeders? both candidate-4.2 and candidate-4.3 feed into fast-4.3 (after we open the minor version upgrades for that release)

I think only candidate-4.3 feeds fast-4.3.  It's just that we add both 4.2 and 4.3 releases to candidate-4.3.  There is a potential use-case for having multiple feeders, but I think it's all of the candidate-4.* channels feeding the version-agnostic `candidate` channel from the proposal in #112.  And that's going to be a bit tricky because of the possibility of implicit edge blocking as discussed in #112, so I haven't dipped into that here.  Can we punt on it and revisit if/when we get around to considering #112?  Or should we hash it out in this PR?

jwforres:
/hold
We need to sync between ART and OTA teams, there are two different channel automation proposals being implemented.

LalatenduMohanty:
The general agreement that OTA wont be creating  the PRs but ART will be creating them.  So if you can  you can do the required changes in the PR , we can merge it.

wking:
/close

Sounds like ART is going to handle filing all promotion PRs

wking:
Rebased, added `filter` (inspired by EUS), and brought it up to speed with master with 14289df -> 641de46.  Also needs #302 and #697 to keep it from complaining about late promotions.

wking:
Rebased around #759 to cover the new 4.8 channels with 2dc94b4 -> 8265e89.

channels/stable-4.5.yaml
```diff
@@ -37,7 +40,9 @@ versions:
 - 4.5.7
 - 4.5.8
 - 4.5.9
+- 4.5.10
```
vrutkovs, line 43: We're retroactively adding those to stable - is this intended?

vrutkovs: intended - https://github.com/openshift/cincinnati-graph-data/commit/845ac8263349184aebbbbb3f54a945124f146956

channels/fast-4.2.yaml
```diff
@@ -1,4 +1,8 @@
 name: fast-4.2
+feeder:
+  name: candidate-4.2
+  delay: PT24H
```
LalatenduMohanty: Right now it is taking few days between candidate and fast channel promotions.

channels/fast-4.5.yaml
```diff
@@ -1,4 +1,8 @@
 name: fast-4.5
+feeder:
+  name: candidate-4.5
+  delay: PT24H
```
LalatenduMohanty: Right now it is taking few days between candidate and fast channel promotions.

channels/stable-4.5.yaml
```diff
@@ -1,4 +1,7 @@
 name: stable-4.5
+feeder:
+  name: fast-4.5
+  delay: PT24H
```
LalatenduMohanty: It should be PT48H as we promote the stable after 48 hours.

LalatenduMohanty (changes requested):

channels/fast-4.5.yaml
LalatenduMohanty: Ah the value got changed in a different commit.

channels/stable-4.5.yaml
LalatenduMohanty: Ah, the value got changed in a different commit.

LalatenduMohanty (approved):
/lgtm

README.md
wking: The most important action is ART publishing the errata, which is when a release is declared supported, regardless of what we do in Cincinnati.  So you could tombstone a release if you didn't want it auto-promoted.  But if the release is only in candidate, you had better also tell ART to not ship the errata.  And if the errata has shipped, tombstoning fast or stable promotions won't drop any supported-ness.

PR #85, opened 2020-02-28
channels/fast-4.2: Promote 4.2.20

thiagoalessio:
Split of https://github.com/openshift/cincinnati-graph-data/pull/67


eparis:
tagging because this already went into 4.2 stable!

PR #86, opened 2020-02-28
channels/fast-4.3: Promote 4.2.20+amd64

thiagoalessio:
Split of https://github.com/openshift/cincinnati-graph-data/pull/67


eparis:
tagging because this already went into 4.2 stable!

PR #87, opened 2020-03-03
Adding information why 4.3.2 and 4.3.3 is not in fast,stable 4.3

LalatenduMohanty:

### DIFF
--- a/channels/fast-4.3.yaml
+++ b/channels/fast-4.3.yaml
@@ -9,3 +9,4 @@ versions:
 - 4.2.20+amd64
 - 4.3.0
 - 4.3.1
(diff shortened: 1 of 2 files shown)


sdodson:
4.3.4 was never built, but should we add it as well with a reference to https://bugzilla.redhat.com/show_bug.cgi?id=1805726 ?

LalatenduMohanty:
> 4.3.4 was never built, but should we add it as well with a reference to https://bugzilla.redhat.com/show_bug.cgi?id=1805726 ?

https://github.com/openshift/cincinnati-graph-data/pull/88/

PR #81, opened 2020-02-26
candidate-4.2: add 4.2.21

thiagoalessio:

PR #103, opened 2020-03-10
Add 4.2.21 to fast channels

joepvd:


LalatenduMohanty:
We had found two issues while adding it to candidate channel
```

    https://bugzilla.redhat.com/show_bug.cgi?id=1734524 marked UpgradeBlocker
    https://bugzilla.redhat.com/show_bug.cgi?id=1809259

```

https://github.com/openshift/cincinnati-graph-data/pull/81

sdodson:
My opinion is that we shouldn't block on either of those.

sdodson:
/hold cancel
:shipit:

PR #104, opened 2020-03-10
Add 4.2.21 to stable

joepvd:
Note, there might be legit concerns to add the upgrade path to `stable-4.3`.

PR #106, opened 2020-03-11
Add 4.2.21 to stable-4.2

sdodson:
Splitting #104 which was adding 4.2.21 into stable-4.3 as well because we'd like more telemetry data regarding 4.2.21 to 4.3 edges.


sdodson:
Wait on #103

eparis:
why not stable-4.3?

eparis:
if I look through git history i'm not going to have any idea why we did this

PR #111, opened 2020-03-12
channels: Add candidate-4.4, fast-4.4, and stable-4.4

wking:
So that Cincinnati doesn't complain when the console bumps its hard-coded list of channels and starts asking for stable-4.4 (openshift/enhancements#123 is in flight to remove the need for the console to hard-code available channels, but hasn't seen much recent progress).  And even without the console issue, the installer has been putting new clusters in stable-4.4 since openshift/installer@d7fb12c07a (openshift/installer#2940).

### DIFF
--- a/channels/candidate-4.4.yaml
+++ b/channels/candidate-4.4.yaml
@@ -1,0 +1,2 @@
+name: candidate-4.4
+versions: []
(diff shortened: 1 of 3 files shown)


sdodson:
/lgtm
We can separate out seeding candidate-4.4 with rc.0

PR #112, opened 2020-03-12
channels: Add version-agnostic candidate, fast, and stable channels

wking:
~For folks who want to flow from one `4.y` to the next `4.(y+1)` and onward without having to manually bump their channel choice.~ [edit: for now, these are internal channels; we may revisit making them public later on]

~I'm seeding these with the current content of the 4.3 channels.  I haven't gone back to consider the `*-4.1` and `*-4.2` channels, because the version-agnostic channels will not just be a blind union.  For example, 4.1.20 is in stable-4.1, but not in stable-4.2. `blocked-edges/4.2.0.yaml` blocks 4.1.20 -> 4.2.0, but without a more careful audit than I've done, I'm not sure if all unsafe edges are represented in blocked-edges (vs. just relying on the membership in the channel files to avoid the unsafe edges).~ [edit: basic approach from this paragraph still applies, but see the individual commit messages for current details]


LalatenduMohanty:
Shouldn't this first vetted by product management?

LalatenduMohanty (dismissed):
Shouldn't this first vetted by product management? Also we should not introduce this in the graph without sufficient consultation with other stake holders and without documentation for this in official OpenShift documentation.

sdodson:
> Shouldn't this first vetted by product management? Also we should not introduce this in the graph without sufficient consultation with other stake holders and without documentation for this in official OpenShift documentation.

Yes, definitely. I like the idea but we need to discuss.
/hold

wking:
Probably want to block this on openshift/enhancements#123 , so the console folks can drop their hard-coded channel list and won't have to worry about channels being added.

wking:
> Probably want to block this on openshift/enhancements#123...

Landed and implemented with openshift/cluster-version-operator#419.

jewzaam:
I really like this, any chance it has legs and may become a real thing?

wking:
I've revived and rerolled this as one possibility for automatically feeding `eus-4.8` with 4.6, 4.7, and 4.8 releases.

sdodson:
/approve
I'm fine with this as long as we coordinate with ART.

We should probably update the PR description though.

sdodson:
/approve
I'm fine with this as long as we coordinate with ART.

We should probably update the PR description though.

wking:
:tada:

/hold cancel

PR #116, opened 2020-03-13
Backfill stable-4.2 nodes into stable-4.3, block 4.3.1

sdodson:
This ensures that those who switched to stable-4.3 with hopes of upgrading to
4.3 but before 4.2.21 will be presented upgrades which will eventually lead to
4.3.5 even though they're on a release that doesn't currently upgrade to 4.3.5.

### DIFF
--- a/blocked-edges/4.3.1.yaml
+++ b/blocked-edges/4.3.1.yaml
@@ -1,0 +1,3 @@
+to: 4.3.1
+from: 4\.2\..*
+# 4.2 -> 4.3 was not deemed stable until 4.3.5
--- a/channels/stable-4.3.yaml
+++ b/channels/stable-4.3.yaml
@@ -1,14 +1,31 @@
 name: stable-4.3
 versions:
-# 4.2.16+amd64 - withheld as it's only valid into 4.3.0,4.3.1 people should go direct to 4.3.5
+- 4.2.0
+- 4.2.1
+- 4.2.2
+# no 4.2.3 because it was rolled into 4.2.4
+- 4.2.4
+# no 4.2.5 because it was cut missing some intended update sources
+# no 4.2.6 because it pulled in CI-built images due to due to release job modification for multi-arch
+- 4.2.7
+- 4.2.8
+- 4.2.9
+- 4.2.10
+# - 4.2.11 failed to run tests, we never officially released it, but we accidentally
+# put it in a channel! (same for s390x)   Now we shouldn't pull it, just in case
+# someone is on it
+- 4.2.11
+- 4.2.12
+- 4.2.13
+- 4.2.14+amd64
+- 4.2.16+amd64
 # not 4.2.17 because we had a long quiet time after 4.2.16 with no releases
-# 4.2.18+amd64 - withheld as it's only valid into 4.3.1, people should go direct to 4.3.5
-# 4.2.19+amd64 - withheld as it's only valid into 4.3.1, people should go direct to 4.3.5
+- 4.2.18+amd64
+- 4.2.19+amd64
 - 4.2.20+amd64
 - 4.2.21+amd64
 # until s390 is released on 4.3 we may not want to include it in 4.3 channels
 # 4.2 -> 4.3 updates occasionally hit RequiredPoolsFailed, fixed in 4.2.18 and rc.0, but not in 4.2.16: https://bugzilla.redhat.com/show_bug.cgi?id=1782152 https://bugzilla.redhat.com/show_bug.cgi?id=1782149
-# not 4.2.17 because we had a long quiet time after 4.2.16 with no releases
 # 4.2.21 edge into stable-4.3 pending further review
 - 4.3.0
 - 4.3.1


channels/stable-4.3.yaml
```diff
@@ -1,9 +1,27 @@
...
+- 4.2.10
+# - 4.2.11 failed to run tests, we never officially released it, but we accidentally
+# put it in a channel! (same for s390x)   Now we shouldn't pull it, just in case
+# someone is on it
+- 4.2.11
+- 4.2.12
+- 4.2.13
+- 4.2.14+amd64
+- 4.2.16+amd64
 # not 4.2.17 because we had a long quiet time after 4.2.16 with no releases
```
wking: We have two of these lines in this file, one from 9f94edc3 and another from 505f7a64.  Can you remove the line further down in this file with this PR?  Or I can file a follow-up PR removing the duplicate line after this PR lands.

channels/stable-4.3.yaml
```diff
@@ -1,9 +1,27 @@
 name: stable-4.3
```
wking: I'm generally not in favor of backfilling (would you include all of stable-4.1 too?), and would rather folks who are this relaxed use version-agnostic channels (#112).  But however we do this, we don't want releases in a stable channel that are not in the corresponding fast channel, etc., so you'll want to add the new releases to `fast-4.3` and `candidate-4.3` as well.

blocked-edges/4.3.1.yaml
```diff
@@ -0,0 +1,4 @@
+to: 4.3.1
+from: 4\.2\..*
+# 4.2 -> 4.3 updates occasionally hit RequiredPoolsFailed, fixed in 4.2.18 and rc.0, but not in 4.2.16: https://bugzilla.redhat.com/show_bug.cgi?id=1782152 https://bugzilla.redhat.com/show_bug.cgi?id=1782149
```
wking: This doesn't apply to 4.3.1, since the 4.2 source is 4.2.18:

```console
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.3.1-x86_64 | grep Upgrade
  Upgrades: 4.2.18, 4.3.0-rc.0, 4.3.0-rc.3, 4.3.0
```

More on that in c641bbdc4 (#60).

blocked-edges/4.3.1.yaml
```diff
@@ -0,0 +1,4 @@
...
+# 4.2 -> 4.3 updates on GCP occasionally have unreachable workloads: https://bugzilla.redhat.com/show_bug.cgi?id=1793635
```
wking: This is still NEW.  If it was grounds for blocking an edge, it would be grounds for blocking all 4.2.* -> 4.3.* edges.  Do we actually have any issues with 4.2.18 -> 4.3.1?  4.3.2 and 4.3.3 block _all_ incoming edges, but I haven't run down the three bugs they cite to see when they were introduced and when they were fixed.

channels/stable-4.3.yaml
```diff
@@ -1,9 +1,27 @@
...
+- 4.2.7
+- 4.2.8
+- 4.2.9
+- 4.2.10
+# - 4.2.11 failed to run tests, we never officially released it, but we accidentally
+# put it in a channel! (same for s390x)   Now we shouldn't pull it, just in case
+# someone is on it
+- 4.2.11
+- 4.2.12
+- 4.2.13
```
wking: When adding these to `candidate-4.3` (which [you should do in this PR][1]), you should also add a `blocked-edges/4.3.0-rc.0` to exclude `.*`, so we don't get a 4.2.13 -> 4.3.0-rc.0 in the candidate graph.  And a `blocked-edges/4.3.0-rc.3` to exclude `4\.2\.16`, because it's easy enough for those folks to flow down 4.2.z and jump off to 4.3 along a later, already stable, edge.  Edges into 4.3:

```console
$ for RC in 0 3; do oc adm release info "quay.io/openshift-release-dev/ocp-release:4.3.0-rc.${RC}-x86_64" | grep Upgrade; done
  Upgrades: 4.2.13
  Upgrades: 4.2.16, 4.3.0-rc.0, 4.3.0-rc.1, 4.3.0-rc.2
$ for z in $(seq 0 5); do oc adm release info "quay.io/openshift-release-dev/ocp-release:4.3.${z}-x86_64" | grep Upgrade; done
  Upgrades: 4.2.16, 4.3.0-rc.0, 4.3.0-rc.1, 4.3.0-rc.2, 4.3.0-rc.3
  Upgrades: 4.2.18, 4.3.0-rc.0, 4.3.0-rc.3, 4.3.0
  Upgrades: 4.2.19, 4.3.0, 4.3.1
  Upgrades: 4.2.20, 4.3.0, 4.3.1, 4.3.2
error: image does not exist
  Upgrades: 4.2.21, 4.2.22, 4.3.0, 4.3.1, 4.3.2, 4.3.3
```

So the other 4.2 sources are all in this stable-4.3 file already.

[1]: https://github.com/openshift/cincinnati-graph-data/pull/116/files#r392028195

channels/stable-4.3.yaml
sdodson: ack

channels/stable-4.3.yaml
sdodson: Yes, I think version agnostic channels would be preferred but that's too big of a change to make without more consideration from the architects team.

The main thing here is we have around 30 clusters in stable-4.3 and have for a few weeks because people've been anticipating a 4.3 upgrade any way now. Most of those clusters are not on 4.2.21 yet so they'd have to go back to stable-4.2 upgrade to 4.2.21 and switch back with out backfilling, correct?

blocked-edges/4.3.1.yaml
sdodson: /shrug
There was a cluster of bugs fixed in 4.3.5 that led us to believe it was the first reliable upgrade path from 4.2.*. I've lost track of specifics and it doesn't warrant digging that deep now in my opinion. I'll update this.

channels/stable-4.3.yaml
sdodson: I'm not that concerned about candidate continuity but I think we should explore adding CI to ensure that `stable-4.x` is a subgraph of `fast-4.x` which is a subgraph of `candidate-4.x`.

sdodson:
/cc @eparis 
please lend us your thoughts on this

channels/stable-4.3.yaml
wking: > Most of those clusters are not on 4.2.21 yet so they'd have to go back to stable-4.2 upgrade to 4.2.21 and switch back with out backfilling, correct?

Yup.  But we alert on `VersionNotFound`, right (I hope ;)?  If we don't, we should.  So folks interested in jumping to 4.3 from earlier 4.2 should have changed their channel to stable-4.3, immediately seen the alert and/or `RetrievedUpdates` go `False`, and then changed back to whatever they were on before.

blocked-edges/4.3.1.yaml
wking: > I've lost track of specifics and it doesn't warrant digging that deep now in my opinion.

This is why we should be documenting the specifics.  I'll see about filing a parallel PR expanding on [these links][1].

[1]: https://github.com/openshift/cincinnati-graph-data/blob/4f4145dc35d44e0eccc4ec3c27d4d0aee0d69e1a/blocked-edges/4.3.3.yaml#L1

channels/stable-4.3.yaml
wking: > I think we should explore adding CI to ensure that stable-4.x is a subgraph of fast-4.x which is a subgraph of candidate-4.x.

There's an open (internal) [ticket][1] for that.

[1]: https://issues.redhat.com/browse/OTA-126

sdodson:
/close
No one seems to be clamoring for this.

PR #117, opened 2020-03-13
Add 4.2.22 to stable

joepvd:


sdodson:
/assign @wking 
/uncc @lucab
For Monday

wking:
Went into the corresponding fast channels [2d ago][1], so we've had enough cook time.  Looking through Telemetry/Insights, there are a few clusters which had update issues, but they seem to have either punched through, had preexisting conditions, or didn't report enough Telemetry/Insights to diagnose.

/lgtm

[1]: https://github.com/openshift/cincinnati-graph-data/pull/107#event-3120375993

PR #115, opened 2020-03-12
Add 4.2.23 to candidate

joepvd:

PR #121, opened 2020-03-16
channels: Promote 4.2.23 to fast-4.2 and fast-4.3

wking:
It was promoted the feeder candidate-4.2 and candidate-4.3 by dc2a3c4cd8 (Merge pull request #115 from joepvd/candidate-4.2.23, 2020-03-13).


LalatenduMohanty:
There is a regression got introduced to 4.2.22 and later builds, see https://github.com/openshift/cincinnati-graph-data/pull/125 , I would suggest that we drop this build and add the build with the fix.

wking:
/hold

We might want to tombstone this instead because of #125.

wking:
I'm just going to close this.  I'll file a tombstoning PR in a bit.

/close

wking:
Tombstoned in #127.

PR #122, opened 2020-03-18
channels: Promote 4.2.23 to stable-4.2 and stable-4.3

thiagoalessio:


wking:
Blocked on #121

LalatenduMohanty:
There is a regression got introduced to 4.2.22 and later builds, see https://github.com/openshift/cincinnati-graph-data/pull/125 , I would suggest that we drop this build and add the build with the fix.

wking:
I'm just going to close this.  I'll file a tombstoning PR in a bit.

/close

wking:
Tombstoned in #127.

PR #118, opened 2020-03-13
add rc.1 in candidate-4.4.yaml

sosiouxme:

PR #123, opened 2020-03-18
Block edges to 4.4.0-rc.1 from 4.3.5

LalatenduMohanty:


wking:
Do we believe these bugs are new in rc.1?  Can you add notes about why they do not affect 4.3 -> rc.0 or rc.0 -> rc.1?

wking:
Wait, with #118 still open, we can just tombstone rc.1 without blocking edges.

/hold

wking:
Maybe reroll to tombstone rc.1?

wking:
#127 tombstones rc.1 so it never lands in any channels.

/close

PR #125, opened 2020-03-18
Block upgrades to 4.3.5 and 4.2.22

marun:
These releases enable automated ca rotation without unique ca serial numbers which can break non-golang tls clients as per [1].

1: https://bugzilla.redhat.com/show_bug.cgi?id=1810036

PR #127, opened 2020-03-18
blocked-edges/4.2.23: Block all incoming edges on the service CA bug 1810036

wking:
Also tombstone affected releases to avoid further channel promotion for affected releases for #125.  Details in the commit message.


wking:
/hold

While I work in the OAuth series for [rhbz#1801573][1].

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1801573

wking:
Updated to include references to the OAuth cert-reload series.

/hold cancel

wking:
Heh, [publish preflight failed][1] on a Quay flake:

```
    meta = get_release_metadata(node=node)
...
urllib2.HTTPError: HTTP Error 500: Internal Server Error
```

/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati-graph-data/127/pull-ci-openshift-cincinnati-graph-data-master-publish/232

sdodson:
> I am still not convinced that we need this much information in the commit message [5d96970](https://github.com/openshift/cincinnati-graph-data/commit/5d96970acc5bedecd529168582cc72ae32b659fc) If anyone else thinks this is necessary feel free to merge this

I don't think it's necessary though I wouldn't block this PR on it. From discussing with the release admins a link to a BZ comment with a properly filled out impact assessment is sufficient to record the reasoning behind blocking of edges. This assumes that the impact assessment clearly defines the versions affected which should be populated by the BZ assginee *not* OTA team.

LalatenduMohanty:
I will move the discussion about the commit message somewhere else.

channels/candidate-4.4.yaml
```diff
@@ -1,4 +1,7 @@
 name: candidate-4.4
 versions:
 - 4.3.5
+# No 4.3.6 because of https://bugzilla.redhat.com/show_bug.cgi?id=1811886
+# No 4.3.7 because of the service CA rotation issue: https://bugzilla.redhat.com/show_bug.cgi?id=1810036 https://bugzilla.redhat.com/show_bug.cgi?id=1801573
 - 4.4.0-rc.0
+# No 4.4.0-rc.1 because of the service CA rotation issue: https://bugzilla.redhat.com/show_bug.cgi?id=1810036
```
LalatenduMohanty: This is a mistake, The 1810036 causes the issue for upgrades but for fresh installs there is no issue for 13months. So it should be in candidate but the edges should be blocked.

wking:
I can move the affected-version information over to the respective bugs...

wking:
I've pushed [a commend to the bug][1] with the workup for "who is impacted?", and pushed 0a35fcd -> 7c0b935 here to link that comment instead of inlining the workup in the commit message.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1810036#c11

channels/candidate-4.4.yaml
wking: > This is a mistake, The 1810036 causes the issue for upgrades but for fresh installs there is no issue for 13months.

I wasn't aware that RCs were hard to cut, but turns out they're ~weekly.  Updated to add 4.4.0-rc.1 with no incoming edges with db2267a53846.

wking:
I've pushed 8e43ba7 dropping candidate-only blocked edges (`rm -f blocked-edges/*rc*`).

blocked-edges/4.2.23.yaml
```diff
@@ -0,0 +1,8 @@
+# Upgrading any release to 4.2.23 will enable automated service ca
+# rotation without unique ca serial numbers. This will result in
+# condition reported in the linked bz, and only be recoverable by
+# manual service ca rotation.
+#
+# Reference: https://bugzilla.redhat.com/show_bug.cgi?id=1810036 https://bugzilla.redhat.com/show_bug.cgi?id=1801573
+to: 4.2.23
+from: .*
```
wking: 4.2.23 is in stable-4.2, etc., so we do want to block the edge on this.

LalatenduMohanty: Yes.

wking:
[publish][1]:

```
could not wait for build: could not create watcher for build src: the server has asked for the client to provide credentials (get builds.build.openshift.io)
```

Probably the CI-cluster outage.

/retest

[1]: https://coreos.slack.com/archives/CFUGK0K9R/p1584641838038700

PR #128, opened 2020-03-19
add 4.4.0-rc.1 to candidate

vfreex:


LalatenduMohanty:
Is 4.4.0-rc1 is a re-build with fix to BZ 1810036 because we already have the PR for it in https://github.com/openshift/cincinnati-graph-data/pull/118 and https://github.com/openshift/cincinnati-graph-data/pull/127 (edited)

sdodson:
/close
#118 already considered this, lets just move on.

PR #129, opened 2020-03-19
add 4.4.0-rc.2 to candidate

vfreex:


LalatenduMohanty:
I am not able to verify that  `4.3 -> 4.4 update hangs on ClusterOperatorDegraded, NodeInstallerDegraded, RevisionControllerDegraded, TargetConfigControllerDegraded -> https://bugzilla.redhat.com/show_bug.cgi?id=1812584 (This is in verified state)` is fixed in this build.

Also following bugs are not fixed in the build.  
* cluster-etcd-operator should not scale when upgrading from 4.3 to 4.4 -> https://bugzilla.redhat.com/show_bug.cgi?id=1813512
* "You are attempting to import a cert with the same issuer/serial as an existing cert, but that is not the same cert" on FIPS enabled cluster after upgrade https://bugzilla.redhat.com/show_bug.cgi?id=1810418

So we are fine introducing this build in candidate but we need to block the edge to this.

wking:
> So we are fine introducing this build in candidate but we need to block the edge to this.

New discussion: we don't care about gating update edges in candidate channels, because the idea is to allow folks to experiment.  Remove all blocked edges that only impact candidate channels.

/lgtm

wking:
Needs a rebase around #127.

sdodson:
Rebased.

PR #135, opened 2020-03-23
Switch logging level to 'info' by default

eparis:
The vast majority of the time we run this script we don't want
the debug data. Only the INFO data.

### DIFF
--- a/hack/graph-util.py
+++ b/hack/graph-util.py
@@ -28,7 +28,6 @@ except ImportError:
 _VERSION_REGEXP = re.compile('^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$')
 logging.basicConfig(format='%(levelname)s: %(message)s')
 _LOGGER = logging.getLogger(__name__)
-_LOGGER.setLevel(logging.DEBUG)
 
 def get_architecture(meta):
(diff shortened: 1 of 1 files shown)


jwforres:
Big plus one, would like @LalatenduMohanty / @wking to ack this won't affect OTA team at all (just cancel the hold if you are ok with it)

/lgtm
/hold

hack/graph-util.py
```diff
@@ -472,18 +471,17 @@ def get_token(args):
     return None
 
 def set_log_level(args):
-    if args.verbose == "info":
-        _LOGGER.setLevel(logging.INFO)
-        return
-    _LOGGER.setLevel(logging.DEBUG)
+    _LOGGER.setLevel(logging.INFO)
+    if args.verbose == "debug":
+        _LOGGER.setLevel(logging.DEBUG)
```
wking: nit: this could be:

```python
_LOGGER.setLevel(args.verbose.upper())
```

without needing a switch on values.  `setLevel` support for string arguments is [new in 3.2][1].  And it works in my Python 2.7.17, although I don't see it [documented for Python 2][2].

[1]: https://docs.python.org/3/library/logging.html#logging.Logger.setLevel
[2]: https://docs.python.org/2/library/logging.html#logging.Logger.setLevel

hack/graph-util.py
```diff
@@ -472,18 +471,17 @@ def get_token(args):
...
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Utilities for managing graph data.')
     parser.add_argument(
         '-v', '--verbose',
```
wking: nit: `--log-level` extends more gracefully if we ever wanted to support `WARNING`, etc.  But since this script is going away soon (we hope), probably doesn't matter either way.

wking:
Left two nits inline.  Doesn't matter to me if you want to address or ignore them.

Before we land this PR, I'd like openshift/release#7867 to land (although it would need a bump if you wanted to [pivot to `--log-level`][1]).

[1]: https://github.com/openshift/cincinnati-graph-data/pull/135#discussion_r396776560

eparis:
(I'd have totally fixed your nits if this was going to stick around)

PR #136, opened 2020-03-23
blocked-edges/4.3: Block all 4.2->4.3 on PodIP vs. PodIPs, bug 1816302

wking:
Seems like a low-probability flake (we've had many successful 4.2->4.3 before seeing this), but we don't have a recovery procedure yet, so pull the edges while we get more details.

https://bugzilla.redhat.com/show_bug.cgi?id=1816302

### DIFF
--- a/blocked-edges/4.3.0.yaml
+++ b/blocked-edges/4.3.0.yaml
@@ -1,5 +1,6 @@
 to: 4.3.0
 from: .*
+# 4.2 -> 4.3 updates occasionally hit PodIP vs. PodIPs, https://bugzilla.redhat.com/show_bug.cgi?id=1816302
 # 4.2 -> 4.3 updates hit TargetDown for cluster-autoscaler-operator, fixed in 4.3.5, https://bugzilla.redhat.com/show_bug.cgi?id=1801300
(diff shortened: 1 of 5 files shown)


wking:
https://bugzilla.redhat.com/show_bug.cgi?id=1816302#c4 has a recovery procedure (deleting a stuck pod).  Not clear to me whether that impacts edge-blocking or not, since we're still not clear on whether this is a regression.

sdodson:
/hold
I'm fine with this sitting edge in candidate right now. We should however firm up our understanding of remediation and the scope of clusters we expect to be affected before we decide what happens with fast-4.3.

LalatenduMohanty:
> I'm fine with this sitting edge in candidate right now. We should however firm up our understanding of remediation and the scope of clusters we expect to be affected before we decide what happens with fast-4.3.

Right, we can remove the blocking edge for 4.3.8 which is in candidate. Other than that we should be blocking the edge. We can come back and add the edges back if we find new information to support it.

LalatenduMohanty:
/lgtm cancel
lets remove the blocking edge for 4.3.8 which is in candidate channel

sdodson:
/hold cancel
/lgtm
Now that 4.3.8 is ready to promote into fast lets go ahead and block 4.2 to 4.3. We can come back and drop this block once we have a better risk assessment.

sdodson:
rebased
/lgtm

PR #146, opened 2020-03-26
blocked-edges/4.3.9: Block 4.2 -> 4.3.9 on PodIP vs. PodIPs, bug 1816302

wking:
Still not fixed in 4.3.z, [rhbz#1817040][1].

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1817040


LalatenduMohanty:
The general agreement is that we need this before going to fast channel. But for candidate we want to keep the edge in the hope that we will get more useful data.

sdodson:
/lgtm
4.3.9 erratum are in push

PR #148, opened 2020-03-31
channels/stable-4.2: Promote 4.1.38 to stable-4.2

wking:
It was promoted the feeder fast-4.2 by 78a82d3016 (#120, 2020-03-24).


wking:
I've taken a look at Telemetry/Insights for this transition.  It's just adding the 4.1.38 -> 4.2.25 edge, and there aren't all that many clusters left on 4.1 so update volume is low.  A few clusters had issues with machine-config going degraded on `RequiredPoolsFailed`, but they seem to have made it through the transition eventually, and there wasn't enough data in Insights to pinpoint a cause (could be due to custom machine setups or some such).  Since we already support this edge via `fast-4.2`, I don't see a problem with promoting it to stable now.

eparis:
/approve cancel
where is stable-4.3?

PR #154, opened 2020-04-01
Add 4.3.10 to fast

joepvd:

PR #155, opened 2020-04-01
Add 4.3.10 to stable

joepvd:


wking:
/hold

Need #154 to land and cook first.

vrutkovs:
5/39 major upgrades got stuck (one case is being investigated, others are mutated sccs)

47 minor upgrades with no critical issue.

I think its good to go
/lgtm
/hold cancel
/cc @eparis

sdodson:
> 5/39 major upgrades got stuck (one case is being investigated, others are mutated sccs)

The mutated SCCs messaging in product seems relatively clear but I've asked the team that introduced this change to post a heads up to lists with customer facing folks so they're aware of it before their customers run into the problem.

PR #156, opened 2020-04-03
hack/errata: Poll for recent errata events

wking:
And push to Slack's Incoming WebHook.

### DIFF
--- a/hack/errata.py
+++ b/hack/errata.py
@@ -60,8 +60,13 @@ def poll(data_grepper='https://datagrepper.engineering.redhat.com/raw', period=N
         params['page'] = page
         uri = '{}?{}'.format(data_grepper, urllib.parse.urlencode(params))
         _LOGGER.debug('query page {}: {}'.format(page, uri))
-        with urllib.request.urlopen(uri) as f:
-            data = json.load(codecs.getreader('utf-8')(f))  # hack: should actually respect Content-Type
+        try:
+            with urllib.request.urlopen(uri) as f:
+                data = json.load(codecs.getreader('utf-8')(f))  # hack: should actually respect Content-Type
+        except Exception as error:
+            _LOGGER.error('{}: {}'.format(uri, error))
+            time.sleep(10)
+            continue
         for raw_message in data['raw_messages']:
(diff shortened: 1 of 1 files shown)


hack/errata.py
```diff
@@ -0,0 +1,100 @@
...
+    with open(path, 'w') as f:
+        json.dump(cache, f, sort_keys=True, indent=2)
+
+
+def run(poll_period=datetime.timedelta(seconds=3600), cache=None, webhook=None, **kwargs):
+    next_time = datetime.datetime.now()
+    while True:
+        _LOGGER.debug('poll for messages')
+        for message in poll(period=2*poll_period, **kwargs):
+            if cache and message['fulladvisory'] in cache or 'bug fix update' not in message['synopsis']:
```
wking, line 37: `bug fix update` is a cheap hack.  We should really be comparing `fulladvisory` like `RHBA-2020:0936-04` to Cincinnati URI like:

```console
$ curl -sH 'Accept:application/json' 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=candidate-4.2' | jq -r '.nodes[] | select(.version == "4.2.26").metadata.url'
https://access.redhat.com/errata/RHBA-2020:0936
```

Although I'm not clear on where the `-04` suffix went.

LalatenduMohanty:
@wking How are you planning to integrate this in our workflow?

LalatenduMohanty:
If this can push a comment to the github PR saying the errata has been shipped will be useful.

wking:
> If this can push a comment to the github PR saying the errata has been shipped will be useful.

It cannot do that yet because:

* No GitHub token yet, although we could create one.
* We'd need logic for the script to identify the appropriate PR, possibly by grepping for `fast` and the version number from the errata synopsis in PR titles?  Possible, but not something I wanted to bite off for v1.

Longer term, yeah, this could comment on PRs, or auto-merge PRs, or file new PRs, or file and merge new PRs, or whatever.  I think we'll start out with "ping us in Slack" and see how that goes.  We can figure out what v2 looks like after we get tired of the manual pings ;).

PR #166, opened 2020-04-10
add 4.3.12 to candidate

vfreex:


eparis:
Don't we put 4.3 candidates into 4.4?

sdodson:
> Don't we put 4.3 candidates into 4.4?

Yes, we should start that now as well as fast-4.4 and stable-4.4 in those PRs. Not sure where to record this so that ART knows to do this going forward.

vfreex:
I didn't notice this change before. Now it is updated.
I will also change our team's procedure docs for awareness.

PR #167, opened 2020-04-10
add 4.3.12 to fast

vfreex:


wking:
/hold

Blocked on #166 and errata publication.

vfreex:
PR updated.

channels/fast-4.4.yaml
```diff
@@ -1,2 +1,3 @@
 name: fast-4.4
-versions: []
+versions:
+- 4.3.12
```
wking, line 3: openshift/cincinnati-graph-data#166 is still open, so this can happen or not without much impact.  If it happens, it's along the lines of the abandoned-for-lack-of-interest #116.

sdodson:
Merge conflicts on our insistence to drop metadata in channel files. I'll rebase.

sdodson:
/assign @jottofar 
This errata is shipping today.

channels/fast-4.4.yaml
eparis: Is there any chance this makes us have to block edges later? I can't think of a reason... So I think I like this PR.

sdodson: Yes we could end up having to block edges later. But edge blocks aren't channel bound so if 4.3.12 to 4.3.13 or 4.3.12 to 4.4.0 were blocked it'd also be blocked here as well.

The primary motivation here is to land releases before console starts letting people choose 4.4 channels so that if someone were to change before we had minor version upgrades we wouldn't orphan them. For instance, I attempted to go from 4.1.0 to 4.2, the biggest upgrades I could take were 4.1.0 to 4.1.14 to 4.1.25 at which point the console let me switch to stable-4.2 and I could no no where.

Maybe I misunderstood the question though.

sdodson:
I've set approved label so this merges, we'd previously agreed that OTA can add items to channels without architect approval but we've not taken action on #173 so that's not possible.

PR #171, opened 2020-04-13
WIP: Block 4.2 -> 4.3.10 for Azure clusters

wking:
We don't know why we started seeing this in 4.3.10, and installer defaults for Azure CPU capacity haven't changed between 4.2 and 4.3, but I'm pushing this up as a mock up of what "block this edge for some types of clusters" might look like.


wking:
/hold

Just to be extra sure that we don't land this mock up ;).

wking:
I've translated from this brief sketch (which didn't generate much interest) into a formal enhancement with openshift/enhancements#426.

/close

wking:
#1056 should be the real deal :)

PR #173, opened 2020-04-14
Drop channels/OWNERS so that it defaults to top level OWNERS

sdodson:
Net effect being that OTA can add to channels without release-admins
Blocks still require release-admins


channels/OWNERS
```diff
@@ -1,10 +0,0 @@
-# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
-# This file just uses aliases defined in OWNERS_ALIASES.
-
-approvers:
-  - release-admins
```
wking, line 5: I'd rather have ART own candidate promotions, and cincinnati-graph-data-approvers only own fast/stable promotions.  But we can't do that until Cincinnati can load channel YAML recursively from `channels/` (which we'd split into `channels/candidate` and `channels/stable` or some such).

wking: Sounds like ART is going to push directly to master, so they don't need to get involved in `OWNERS`.  In that case, I'm fine with this PR as it stands.

LalatenduMohanty: We can come back to it and add ART to the list of approvers when we have ART ready to push to the candidates.

sdodson:
> /hold for release-admins to approve

Until this merges they have to approve this anyway. I'm going to clear the hold so it pops up on their radar.
/hold cancel

PR #177, opened 2020-04-15
Add 4.2.29 to candidate

thiagoalessio:

PR #178, opened 2020-04-15
Add 4.2.29 to fast

thiagoalessio:


wking:
/hold

#177 is in the merge queue, but this promotion is blocked on errata.

wking:
We want this so we still have supported 4.2 -> 4.3 edges after #184 lands.  Checking Telemetry, there are no concerning updates into or out of 4.2.29, so:

/hold cancel
/lgtm

PR #179, opened 2020-04-15
Add 4.2.29 to stable

thiagoalessio:


wking:
/hold

Blocked on #178 and cook time.

wking:
We want this so we still have stable 4.2 -> 4.3 edges after #184 lands. Checking Telemetry, there are no concerning updates into or out of 4.2.29, and #178 has landed, so:

/hold cancel
/lgtm

PR #180, opened 2020-04-16
4.1.41 prerelease and candidate

shiywang:


wking:
Some of [the update CI][1] failed, but CI has been really struggling today for other reasons, so I'm not going to block on that.  We've also asked for "why were versions skipped" comments [for a while][2], but when I asked @eparis about whether he wanted to hold us to that he didn't respond.  So I'm just going to land this as it stands and see if there's any pushback ;).

/lgtm
/retest

[1]: https://openshift-release.svc.ci.openshift.org/releasestream/4-stable/release/4.1.41#upgrades-from
[2]: https://github.com/openshift/cincinnati-graph-data/commit/ff4c10a8cbe7c04663a5bd731d99f0e0d9dc3d9d#diff-04c6e90faac2675aa89e2176d2eec7d8R24

PR #183, opened 2020-04-16
Add 4.1.20+ to 4.2 channels

sdodson:
The console was updated in 4.1.20 to include the 4.2 channels. We should make sure that we've not abandoned anyone who switched channels after that point.

Ref: https://access.redhat.com/errata/RHBA-2019:3004
https://bugzilla.redhat.com/show_bug.cgi?id=1758636


sdodson:
Not sure how 4.1.25 ended up missing in the middle, but when I attempted to go from 4.1.0 to 4.2 I just took the biggest steps I could on stable-4.1 channel and ended up with 4.1.0 -> 4.1.14 -> 4.1.25 at which point I could switch channels and I was deserted on an island.

sdodson:
/uncc
/cc @wking 
PTAL

sdodson:
I also reviewed 4.2 into 4.3 channels, that's all good. Went into the 4.2 console in 4.2.16.
ref:
https://bugzilla.redhat.com/show_bug.cgi?id=1791182
https://access.redhat.com/errata/RHBA-2020:0107

channels/stable-4.2.yaml
```diff
@@ -1,6 +1,11 @@
 name: stable-4.2
 versions:
+- 4.1.20
+- 4.1.21
+- 4.1.22
+- 4.1.23
```
wking, line 6: This can't go into stable without also going into fast.

wking: /hold

wking:
Checking the edges that this will enable:

```console
$ yaml2json <channels/candidate-4.2.yaml | jq -r '.versions[]' | grep -v 's390x\|-rc' | sed -n 's/^4\.2\.//p' | while read Z; do echo -n "4.2.${Z} "; if test "${Z}" -ge 14; then ARCH=-x86_64; else ARCH=; fi; oc adm release info "quay.io/openshift-release-dev/ocp-release:4.2.${Z}${ARCH}" | grep Upgrade; done | grep ' 4\.1\.2[01235]'
4.2.0   Upgrades: 4.1.20, 4.2.0-rc.4, 4.2.0-rc.5
4.2.1   Upgrades: 4.1.20, 4.1.21, 4.2.0-rc.4, 4.2.0-rc.5
4.2.2   Upgrades: 4.1.21, 4.2.0, 4.2.1
4.2.4   Upgrades: 4.1.22, 4.1.23, 4.2.2
```

And those are already blocked:

```console
$ grep '^from:' blocked-edges/4.2.[0124].yaml
blocked-edges/4.2.0.yaml:from: 4\.1\..*
blocked-edges/4.2.1.yaml:from: .*
blocked-edges/4.2.2.yaml:from: 4\.1\..*
blocked-edges/4.2.4.yaml:from: 4\.1\..*
```

So looks good to me once [fast gets filled in][1].

[1]: https://github.com/openshift/cincinnati-graph-data/pull/183#discussion_r409897814

channels/stable-4.2.yaml
sdodson: ack, thanks

wking:
/lgtm
/hold cancel

Issue was [addressed][1].

[1]: https://github.com/openshift/cincinnati-graph-data/pull/183#discussion_r409907620

PR #184, opened 2020-04-16
Block edges to 4.3.8 through 4.3.12 on DefaultSecurityContextConstraints_Mutated

sdodson:
4.3.8 introduced a change which blocks upgrades when the default SCCs are
mutated. That change was merged under the assumption that the CVO would not
block z-stream upgrades on that condition. However that change had not
merged yet and thus ~40% of clusters which have upgraded to 4.3.8-4.3.12
are now blocked from upgrading with out specifying the --force flag. 4.3.13
relaxes the Upgradeable=False such that you can now apply z-stream updates
when that condition is set. A future update, likely 4.3.14 will revert the change
which began setting that condition.

Because the slim majority of subscribed clusters had already upgraded into
this problem and because of the numerous previously blocked upgrades we
had decided to live with the problem until 4.3.13 ships. So once 4.3.13 is
promoted into stable-4.3 we should stop additional clusters from falling
into the problem.

Reference https://bugzilla.redhat.com/show_bug.cgi?id=1821905#c22

### DIFF
--- a/blocked-edges/4.3.10.yaml
+++ b/blocked-edges/4.3.10.yaml
@@ -1,4 +1,3 @@
-to: 4.3.8
+to: 4.3.10
 from: .*
-# 4.2 -> 4.3 updates occasionally hit PodIP vs. PodIPs, https://bugzilla.redhat.com/show_bug.cgi?id=1816302
 # 4.3.8 - 4.3.12 block updates on mutated SCCs at the intersection of https://bugzilla.redhat.com/show_bug.cgi?id=1821905 https://bugzilla.redhat.com/show_bug.cgi?id=1820231
--- a/blocked-edges/4.3.12.yaml
+++ b/blocked-edges/4.3.12.yaml
@@ -1,4 +1,3 @@
-to: 4.3.8
+to: 4.3.12
 from: .*
-# 4.2 -> 4.3 updates occasionally hit PodIP vs. PodIPs, https://bugzilla.redhat.com/show_bug.cgi?id=1816302
 # 4.3.8 - 4.3.12 block updates on mutated SCCs at the intersection of https://bugzilla.redhat.com/show_bug.cgi?id=1821905 https://bugzilla.redhat.com/show_bug.cgi?id=1820231
--- a/blocked-edges/4.3.11.yaml
+++ b/blocked-edges/4.3.11.yaml
@@ -4,3 +4,4 @@ from: .*
 # Hence the images are rebuilt as 4.3.12. However 4.3.12 release image does not have metadata to update from 4.3.11.
 # 4.3.11 is only released in candidate-4.3 so we should stop edges to it as there are no outgoing edges available for now.
 # BugZilla: https://bugzilla.redhat.com/show_bug.cgi?id=1823378
+# 4.3.8 - 4.3.12 block updates on mutated SCCs at the intersection of https://bugzilla.redhat.com/show_bug.cgi?id=1821905 https://bugzilla.redhat.com/show_bug.cgi?id=1820231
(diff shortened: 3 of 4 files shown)


sdodson:
/hold
For https://github.com/openshift/cincinnati-graph-data/pull/176

sdodson:
Hmm, this will create a gap where 4.3.0 cannot upgrade to 4.3.13. Need to consider what to do there.

sdodson:
This will also raise the minimum version to upgrade from 4.2 to 4.3 from 4.2.21 to 4.2.29 which is undesirable but not unworkable. Which currently moves us from ~ 26% of the 4.2 fleet to < 1%. This could be mitigated by ensuring that 4.3.14 includes upgrades back to 4.2.21 but this seems like we're stretching and would introduce additional confusion.

wking:
> This will also raise the minimum version to upgrade from 4.2 to 4.3 from 4.2.21 to 4.2.29 which is undesirable but not unworkable.

Should just be 4.2.{current} -> 4.2.29 -> 4.3, which doesn't seem that bad?  Even from 4.2.0, it would be 4.2.0 -> 4.2.13 -> 4.2.21 -> 4.2.29 -> 4.3, so still just one extra hop.

wking:
> Hmm, this will create a gap where 4.3.0 cannot upgrade to 4.3.13.

So we should include 4.3.0 in 4.3.14.

blocked-edges/4.3.10.yaml
```diff
@@ -0,0 +1,7 @@
+to: 4.3.10
+from: .*
+# 4.3.11 container images were built multi-arch, but we have not started releasing multi-arch images for 4.3.x yet.
+# Hence the images are rebuilt as 4.3.12. However 4.3.12 release image does not have metadata to update from 4.3.11.
+# 4.3.11 is only released in candidate-4.3 so we should stop edges to it as there are no outgoing edges available for now.
+# BugZilla: https://bugzilla.redhat.com/show_bug.cgi?id=1823378
```
wking, line 6: These 4.3.11 reasons do not apply to *->4.3.10, which you are blocking here.

blocked-edges/4.3.12.yaml
```diff
@@ -0,0 +1,7 @@
+to: 4.3.12
+from: .*
+# 4.3.11 container images were built multi-arch, but we have not started releasing multi-arch images for 4.3.x yet.
+# Hence the images are rebuilt as 4.3.12. However 4.3.12 release image does not have metadata to update from 4.3.11.
+# 4.3.11 is only released in candidate-4.3 so we should stop edges to it as there are no outgoing edges available for now.
+# BugZilla: https://bugzilla.redhat.com/show_bug.cgi?id=1823378
```
wking, line 6: These 4.3.11 reasons do not apply to *->4.3.12, which you are blocking here.

wking:
/lgtm

Just need to wait on #176, or as close to that landing as we feel we need to get so we can land this without ruffling feathers about "where did my update go?".  I'm personally fine landing this today, even if lots of folks are already up in the affected versions, because waiting a few days for #176 to land doesn't seem like a big deal.  But I also understand that 4.2 -> 4.3 connectivity has not been great so far, and that there are clusters out there who have not mutated their default MCCs.  Speaking of which:

/retitle Block edges to 4.3.8 through 4.3.12 on DefaultSecurityContextConstraints_Mutated

should make it a bit easier to find this in the Git logs.

sdodson:
Still holding on this until we work through some decisions internally.

sdodson:
/hold cancel
This is ready to go now.

PR #233, opened 2020-05-07
version: Begin versioning the schema

wking:
So that Cincinnati and other consumers can identify and fail fast if they are fed a schema they will not understand.


README.md
```diff
@@ -13,6 +13,13 @@ The [contributing documentation](CONTRIBUTING.md) covers licencing and the usual
 
 [Cincinnati][] is configured to track the master branch, so it will automatically react to updates made to this repository.
 
+### Schema version
+
+The layout of this repository is versioned via [a `version` file](version), which contains the [Semantic Version][semver] of the schema.
```
LalatenduMohanty, line 18: This is a good idea, +1.

PR #240, opened 2020-05-08
Add 4.4.3 but remove 4.3.18 from stable-4.4

sdodson:
Because we cannot block edges by channel the only way to prevent 4.3 to
4.4 upgrades in the stable channel but prevent VersionNotFound errors
is to temporarilly remove 4.3 versions. Once we intend to promote the 4.3
to 4.4 edge to stable we will re-introduce these 4.3 versions. 4.3.12 and
4.3.13 are safe but nothing 4.3.18+ should be added since it includes edges
to 4.4. This does have the side effect of preventing people who had switched
channels prematurely from getting 4.3.z updates but hopefully this is short lived.

Supersedes #228

### DIFF
--- a/channels/stable-4.4.yaml
+++ b/channels/stable-4.4.yaml
@@ -1,6 +1,12 @@
 name: stable-4.4
 versions:
+# Because we cannot block edges by channel the only way to prevent 4.3 to 4.4
+# upgrades in the stable channel but prevent errors from being thrown is to
+# temporarilly remove 4.3 versions. Once we intend to promote the 4.3 to 4.4
+# edge to stable we will re-introduce these 4.3 versions. 4.3.12 and 4.3.13
+# are safe but nothing 4.3.18+ should be added since it includes edges to 4.4
 - 4.3.12
 - 4.3.13
 # No 4.3.14 because it was missing version which 4.3.14 can be upgraded. It is rebuilt as 4.3.15
-- 4.3.18
+# - 4.3.18 temporarily to prevent 4.3 to 4.4.0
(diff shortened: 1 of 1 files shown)


sdodson:
/hold
/cc @wking @LalatenduMohanty 
PTAL

channels/stable-4.4.yaml
```diff
@@ -1,6 +1,12 @@
 name: stable-4.4
 versions:
+# Because we cannot block edges by channel the only way to prevent 4.3 to 4.4
+# upgrades in the stable channel but prevent errors from being thrown is to
+# temporarilly remove 4.3 versions. Once we intend to promote the 4.3 to 4.4
```
wking, line 5: nit: "temporarilly" -> "temporarily".

LalatenduMohanty:
I am not in favor of removing the 4.3.18 from the channel as it would mean that we are treating fast and stable differently with respect to release stabiliy. If we do not think edge has enough stability we should just remove it.

pamoedom:
Hi Scott, my 2 cents:

How about just adding the version 4.4.3 but without any edge instead of removing 4.3.x nodes from stable-4.4 channel? like an isolated node, I have seen this in the past in 4.3 channels when Engineering had to temporarily disable some upgrade paths.

Best Regards.

sdodson:
@pamoedom That's not possible with the current tooling.

We'll just wait to see what's up with 4.4.4 and then evaluate when that's ready.
/close

pamoedom:
ACK Scott, thanks.

PR #242, opened 2020-05-11
blocked-edges/4.4.3: Block 4.3 -> 4.4.3 on etcd DNS SAN peer certs

wking:
There's [a reasonably easy workaround][1] for folks who do get stuck, so not worth pulling the edge immediately.  But we want to pull the edge once there's a 4.3 -> 4.4 route that is not exposed to the fix, so cluster admins don't have to discover the workaround.  Blocking on that 4.3 -> 4.4 route (e.g. via 4.4.4).

/hold

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1830409#c7

### DIFF
--- a/blocked-edges/4.4.3.yaml
+++ b/blocked-edges/4.4.3.yaml
@@ -1,0 +1,3 @@
+to: 4.4.3
+from: 4\.3\..*
+# 4.3 -> 4.4 updates occasionally stick on etcd NodeInstaller_InstallerPodFailed, https://bugzilla.redhat.com/show_bug.cgi?id=1830510


blocked-edges/4.4.3.yaml
```diff
@@ -0,0 +1,3 @@
+to: 4.4.3
+from: 4\.3\..*
+# 4.3 -> 4.4 updates occasionally stick on etcd NodeInstaller_InstallerPodFailed, https://bugzilla.redhat.com/show_bug.cgi?id=1830409
```
LalatenduMohanty, line 3: The bug link should be https://bugzilla.redhat.com/show_bug.cgi?id=1830789 as it is targeted for 4.4.z

sdodson: Can we update this to reference https://bugzilla.redhat.com/show_bug.cgi?id=1830510, that's the bug that's fixed in 4.4.4 that we feel at least reduces the risk substantially.

wking: I think we link the master-most bug, so we can copy/paste the message into other z streams.  And also because that's where the impact statement should live.  I can update to link the split-off bug.

LalatenduMohanty: @wking I did not get that. Can you update the PR and then I can take a look.

wking: I'm confused by rhbz#1830510, and have [asked for an impact statement](https://bugzilla.redhat.com/show_bug.cgi?id=1830510#c6).

wking: There's an impact statement in rhbz#1830510.  Rerolled to cite that with 7e74f2f -> 5f1b385.

wking:
4.3 -> 4.4.4 is available in fast-4.4 since #251 

/hold cancel

sdodson:
/lgtm
@eparis ptal

wking:
/override ci/prow/publish

Quay is having issues today; don't worry about it.

PR #243, opened 2020-05-13
[WIP] Removing extra entries for the same version from channel yaml files

LalatenduMohanty:
Need to fix entries from some more PRs once they are merged
[] https://github.com/openshift/cincinnati-graph-data/pull/238#discussion_r424370016
[] https://github.com/openshift/cincinnati-graph-data/pull/239/files#r424383526
[] https://github.com/openshift/cincinnati-graph-data/pull/236/files#r424383356


channels/fast-4.4.yaml
```diff
@@ -2,8 +2,6 @@ name: fast-4.4
 versions:
 - 4.3.19
```
steveeJ, line 4: Is this empty line intended?

LalatenduMohanty: Art usually adding empty lines after each entry. They can clarify. cc  @jupierce

PR #253, opened 2020-05-20
Block 4.3.12, 4.3.13 to 4.4.4

sdodson:
Need to have fixes for https://bugzilla.redhat.com/show_bug.cgi?id=1827337
before upgrading to 4.4. We didn't run into this with 4.4.3 because it only included 4.3.18+


sdodson:
We need to make sure that 4.4.5 either has blocks or is built with 4.3.18 as minimum. The former probably the easiest.

sdodson:
@eparis PTAL

PR #285, opened 2020-06-15
blocked-edges/4.4.*: Block 4.3.12 and 4.3.13 -> 4.4

wking:
Extending 0a65fffd68 (#253) to subsequent releases which included the older 4.3 in their baked-in update sources.

Also consolidate the inline comment to be a one-liner for easier
grepping, and drop the unnecessary pipe from [23] character class.


sdodson:
/lgtm
/assign @eparis 
PTAL, we have a hard floor of 4.3.18 for 4.3 to 4.4 upgrades due to the referenced bug.

PR #328, opened 2020-07-13
Blocking edges to 4.5.1

LalatenduMohanty:


blocked-edges/4.5.1.yaml
```diff
@@ -0,0 +1,3 @@
+to: 4.5.1
+from: *
```
sdodson, line 2: All the other examples are `from: .*` i don't think this is a glob.

sdodson: ex https://github.com/openshift/cincinnati-graph-data/blob/master/blocked-edges/4.1.1.yaml

LalatenduMohanty: Good catch, thanks

sdodson:
@eparis @jwforres you're up for approval

PR #337, opened 2020-07-15
block updates to 4.3.29 for s390x

sosiouxme:
Not sure this is the right format but trying to capture that 4.3.29 is broken for one arch.

https://bugzilla.redhat.com/show_bug.cgi?id=1855807 => https://coreos.slack.com/archives/CFFJUNP6C/p1594411217405400

### DIFF
--- a/blocked-edges/4.3.29.yaml
+++ b/blocked-edges/4.3.29.yaml
@@ -1,0 +1,3 @@
+to: 4.3.29+s390x
+from: .*
+# https://bugzilla.redhat.com/show_bug.cgi?id=1855807 (cri-o failure at boot)
--- a/channels/stable-4.3.yaml
+++ b/channels/stable-4.3.yaml
@@ -57,3 +57,5 @@ versions:
 - 4.3.27
 
 - 4.3.28
+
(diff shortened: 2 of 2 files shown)


blocked-edges/4.3.29.yaml
```diff
@@ -0,0 +1,3 @@
+to: 4.3.29+s390x
```
vrutkovs, line 1:
```suggestion
to: 4.3.29-s390x
```
According to https://github.com/openshift/cincinnati-graph-data/blob/master/blocked-edges/4.2.13-s390x.yaml

sosiouxme: i believe `-s390x` was part of the release name in those days. it's not any more...

vrutkovs: Turns out Cincinnati [supports both](https://github.com/openshift/cincinnati/blob/0edcfea04f35693bd11a4429586c1fa8588f3b66/cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs#L322), but `+` would be better

PR #338, opened 2020-07-16
blocked-edges/4.3.29+ppc64le: Block on rhbz#1855807 (CRI-O boot trouble)

wking:
Andy McCrae has confirmed that this bug impacts ppc64le as well ([rhbz#1855807c8][1], sorry, internal link :/).

I think blocking edges in this case is sufficient and the fact that we have promoted the release to fast for these arches means we are committed to promoting the release (but not their incoming edges) to stable as well.  But to minimize the change in this commit, I'm punting on that for now and just following the precedent established by 49b990f3f1 (#337).

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1855807#c8


sosiouxme:
/lgtm
:)

wking:
/override ci/prow/publish

[publish][1] got past its sanity checks and into dummy-POSTs before Quay 429ed it.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-graph-data/338/pull-ci-openshift-cincinnati-graph-data-master-publish/1283888350922543104

PR #356, opened 2020-07-31
Block 4.4.12 to 4.5.4 upgrades

sdodson:


sdodson:
/assign @vrutkovs 
PTAL

sdodson:
Our testing shows a high rate of success when 4.4.13 is the minimum version required to upgrade to 4.5.

PR #378, opened 2020-08-07
Adding 4.4.z edges to 4.5 stable channels

LalatenduMohanty:
The nodes are same as 4.5 fast channel.


channels/stable-4.5.yaml
```diff
@@ -1,6 +1,20 @@
 name: stable-4.5
 versions:
 
+- 4.4.16
```
sdodson, line 4: lets drop this for now since #372 hasn't merged yet

LalatenduMohanty: Fixed it.

PR #422, opened 2020-09-03
errata.py: autoLGTM the fast PR after errata has shipped

vrutkovs:
This change would add a comment with the LGTM to the PR promoting to
fast after errata has changed status to SHIPPED_LIVE. After datagrepper
gets notified that errata state has changed, it would parse PRs
targetting fast (expects default title), extract errata number and
compare to the received message. If errata numbers match the script
would comment with `/lgtm`.

This requires the hook to be deployed with GITHUB_TOKEN set, which
would be taken care of in internal autoreport repo.

/cc @LalatenduMohanty @wking

### DIFF
--- a/hack/errata.py
+++ b/hack/errata.py
@@ -12,6 +12,8 @@ import time
 import urllib.parse
 import urllib.request
 
+import github
+
 
 logging.basicConfig(level=logging.DEBUG)
 _LOGGER = logging.getLogger()
@@ -30,13 +32,19 @@ def save(path, cache):
         json.dump(cache, f, sort_keys=True, indent=2)
 
 
-def run(poll_period=datetime.timedelta(seconds=3600), cache=None, webhook=None, **kwargs):
+def run(poll_period=datetime.timedelta(seconds=3600),
+        cache=None,
+        webhook=None,
+        githubrepo=None,
+        githubtoken=None,
+        **kwargs):
     next_time = datetime.datetime.now()
     while True:
         _LOGGER.debug('poll for messages')
         for message in poll(period=2*poll_period, **kwargs):
             if cache and message['fulladvisory'] in cache or 'bug fix update' not in message['synopsis']:
                 continue
+            message['approved_pr'] = lgtm_fast_pr_for_errata(githubrepo, githubtoken, message)
             notify(message=message, webhook=webhook)
             if cache is not None:
                 cache[message['fulladvisory']] = {
@@ -83,11 +91,73 @@ def notify(message, webhook=None):
         print(message)
         return
 
+    msg_text = '<!subteam^STE7S7ZU2>: {fulladvisory} shipped {when}: {synopsis}'.format(**message)
+    if  message['approved_pr']:
+        msg_text += "\nPR {approved_pr} has been approved".format(**message)
+
     urllib.request.urlopen(webhook, data=urllib.parse.urlencode({
         'payload': {
-            'text': '<!subteam^STE7S7ZU2>: {fulladvisory} shipped {when}: {synopsis}'.format(**message),
+            'text': msg_text,
         },
     }).encode('utf-8'))
+
+
+def get_open_prs_to_fast(repo):
+    query_params = {
+        'state': 'open',
+        'base': 'master',
+        'author': 'openshift-bot',
+        'sort': 'created',
+    }
+    for pr in repo.get_pulls(**query_params):
+        try:
+            # Skip unknown PRs
+            if not pr.title.startswith("Enable "):
+                continue
+            # Ignore PRs which don't target fast
+            if pr.title.split(" ")[3] != "fast":
+                continue
+            yield pr
+        except Exception as e:
+            _LOGGER.warn("Failed to parse {}: {}".format(pr.number, e))
+
+
+def extract_errata_number_from_body(body):
+    ERRATA_MARKER = 'https://errata.devel.redhat.com/advisory/'
+    first_line = body.split('\n')[0]
+    links = [
+        x for x in first_line.split(' ') if x.startswith(ERRATA_MARKER)
+    ]
+    if len(links) == 0:
+        _LOGGER.warn("No links found in PR body: {}".format(body))
+        return None
+    errata_num = links[0].rsplit('/', 1)[-1]
+
+    try:
+        return int(errata_num)
+    except ValueError:
+        _LOGGER.warn("Failed to convert PR number to int: {}".format(errata_num))
+        return None
+
+
+def lgtm_fast_pr_for_errata(githubrepo, githubtoken, message):
+    if not githubtoken:
+        _LOGGER.debug("Skipping fast PR check: no github token set")
+        return
+
+    github_object = github.Github(githubtoken)
+    repo = github_object.get_repo(githubrepo)
+
+    for pr in get_open_prs_to_fast(repo):
+        errata_num = extract_errata_number_from_body(pr.body)
+        if not errata_num or errata_num != message.get('errata_id'):
+            continue
+
+        _LOGGER.debug("Found PR #{} promoting to fast for {}".format(pr.number, errata_num))
+        msg = "Autoapproving PR to fast after the errata has shipped\n/lgtm"
+        pr.create_issue_comment(msg)
+        _LOGGER.debug("Commented in {}".format(pr.url))
+        return pr.url
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


hack/errata.py
```diff
@@ -38,6 +39,7 @@ def run(poll_period=datetime.timedelta(seconds=3600), cache=None, webhook=None,
             if cache and message['fulladvisory'] in cache or 'bug fix update' not in message['synopsis']:
                 continue
             notify(message=message, webhook=webhook)
+            lgtm_fast_pr_for_errata(message)
```
wking, line 42: Can we shift this up, return any approved PR, and mention it in the `notify` message?

hack/errata.py
```diff
@@ -90,6 +92,58 @@ def notify(message, webhook=None):
...
+        try:
+            # Skip unknown PRs
+            if not pr.title.startswith("Enable "):
+                continue
+            # Ignore PRs which don't target fast
+            if pr.title.split(" ")[3] != "fast":
+                continue
+            yield pr.number, pr.body
+        except Exception as e:
+            print(f"Failed to parse {pr.number}: {e}")
```
wking, line 107: My old RHEL still doesn't support f-strings.  Can we use `.format(...)`?

hack/errata.py
```diff
@@ -90,6 +92,58 @@ def notify(message, webhook=None):
     }).encode('utf-8'))
 
 
+def get_open_prs_to_fast(repo):
+    pulls = repo.get_pulls(state='open', sort='created', base='master')
+    for pr in pulls:
```
wking, line 97: Nit: you don't save much by defining `pulls`.  Maybe `for pull in repo.get_pulls(...):`?

hack/errata.py
```diff
@@ -90,6 +92,58 @@ def notify(message, webhook=None):
...
+        try:
+            # Skip unknown PRs
+            if not pr.title.startswith("Enable "):
+                continue
+            # Ignore PRs which don't target fast
+            if pr.title.split(" ")[3] != "fast":
+                continue
+            yield pr.number, pr.body
```
wking, line 105: This makes the order an implicit API.  Can we `yield pr` (or `yield pull`, if you pivot from `pr` to `pull` for consistency with `get_pulls`)?

hack/errata.py
```diff
@@ -90,6 +92,58 @@ def notify(message, webhook=None):
...
+        except Exception as e:
```
wking, line 106: Nit: the `yield` does not need to be covered by this parsing-focused try/except.

hack/errata.py
```diff
@@ -90,6 +92,58 @@ def notify(message, webhook=None):
...
+
+def extract_errata_number_from_body(body):
+    ERRATA_MARKER = 'https://errata.devel.redhat.com/advisory/'
+    first_line = body.split('\n')[0]
+    links = [
+        x for x in first_line.split(' ') if x.startswith(ERRATA_MARKER)
+    ]
+    if len(links) == 0:
+        return None
+    errata_num = links[0].split('/')[-1]
```
wking, line 118: nit: `rsplit('/', 1)`?

hack/errata.py
wking: Also, we can import `logging` for these warning messages, but I can do that in follow-up work if you prefer.

hack/errata.py
```diff
@@ -90,6 +92,58 @@ def notify(message, webhook=None):
...
+
+    try:
+        return int(errata_num)
+    except ValueError:
+        return None
```
wking, line 123: Log this ignored error?

hack/errata.py
```diff
@@ -90,6 +92,58 @@ def notify(message, webhook=None):
...
+
+
+def lgtm_fast_pr_for_errata(message):
+    github_token = os.environ.get("GITHUB_TOKEN")
+    if not github_token:
+        print(f"Skipping fast PR check: no github token set")
+        return
+
+    github_object = Github(github_token)
+    repo = github_object.get_repo("openshift/cincinnati-graph-data")
```
wking, line 133: Nit: Add argparse options for this repo and the token, so they are less magical and more of an explicit config.

hack/errata.py
vrutkovs: Included URL in the notification - 81af18437430ec02fb94bc314ae57c0f803d3c83

hack/errata.py
```diff
@@ -11,6 +11,7 @@
 import time
 import urllib.parse
 import urllib.request
+from github import Github
```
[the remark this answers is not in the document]
vrutkovs: Sure, lets keep those uniform. Fixed in 67346cb17ecc109984f8f4a8ec46e73b4d3bb5e4

hack/errata.py
vrutkovs: we don't want to yield if exception occurred

hack/errata.py
vrutkovs: Fixed f-strings in 52ff346fe5975f072c7ca0d2c1e704e7708c9ad3

hack/errata.py
vrutkovs: Fixed in 212d2366324e63f49de66ca5c740fa9c162478c8

hack/errata.py
vrutkovs: Added more logging in 212d2366324e63f49de66ca5c740fa9c162478c8

hack/errata.py
```diff
@@ -83,13 +90,70 @@ def notify(message, webhook=None):
...
+
+    try:
+        return int(errata_num)
+    except ValueError:
+        print("Failed to convert PR number to int: {}".format(errata_num))
+        return None
+
+
+def lgtm_fast_pr_for_errata(githubrepo, message):
+    github_token = os.environ.get("GITHUB_TOKEN")
```
wking, line 137: Can we pass this in from an argparse arg too, following the `webhook` pattern?

hack/errata.py
```diff
@@ -83,13 +90,70 @@ def notify(message, webhook=None):
...
+
+
+def extract_errata_number_from_body(body):
+    ERRATA_MARKER = 'https://errata.devel.redhat.com/advisory/'
+    first_line = body.split('\n')[0]
+    links = [
+        x for x in first_line.split(' ') if x.startswith(ERRATA_MARKER)
+    ]
+    if len(links) == 0:
+        print("No links found in PR body: {}".format(body))
```
wking, line 125: s/print/_LOGGER.warning/ for your diff.  Only `print` should be the no-webhook notification mode.

hack/errata.py
```diff
@@ -83,13 +90,70 @@ def notify(message, webhook=None):
...
     urllib.request.urlopen(webhook, data=urllib.parse.urlencode({
         'payload': {
-            'text': '<!subteam^STE7S7ZU2>: {fulladvisory} shipped {when}: {synopsis}'.format(**message),
+            'text': msg_text,
         },
     }).encode('utf-8'))
 
 
+def get_open_prs_to_fast(repo):
+    for pr in repo.get_pulls(state='open', sort='created', base='master'):
```
wking, line 105: Hmm, can we require `openshift-bot` as the author too?

hack/errata.py
vrutkovs: removed remaining f-string in 51a8b4a626c0d33248436ba521be14b2fa5dd976

hack/errata.py
vrutkovs: Fixed in 0c648c9d184320b39b668223eb69fdf7e890ded4

hack/errata.py
vrutkovs: Fixed in 8a1eb6a3401957bc0e83372ceb9ad673f851826a

wking:
Looks good to me.  Squash, and we'll see what @LalatenduMohanty has to say?

hack/errata.py
```diff
@@ -83,13 +91,75 @@ def notify(message, webhook=None):
...
+    github_object = github.Github(githubtoken)
+    repo = github_object.get_repo(githubrepo)
+
+    for pr in get_open_prs_to_fast(repo):
+        errata_num = extract_errata_number_from_body(pr.body)
+        if not errata_num or errata_num != message.get('errata_id'):
+            continue
+
+        _LOGGER.debug("Found PR #{} promoting to fast for {}".format(pr.number, errata_num))
+        msg = "Autoapproving PR to fast after the errata has shipped\n/lgtm"
```
wking, line 157: @LalatenduMohanty points out that we need `/hold cancel` here too (because the candidate promotion has presumably merged long before).  Or we can teach the ART tooling to stop setting holds on the fast PRs, if we want a way to block a PR from auto-merging for whatever reason.

LalatenduMohanty:
/hold

We can not do it without looking in to the success % of the upgrades for the version. Also we should just try to do phased rollout stuff soon which will make this useful and eliminate the downsides. Discussing over slack with Trevor.

LalatenduMohanty:
* https://github.com/openshift/cincinnati-graph-data/pull/422 should not remove the /hold for now.
* If we observe for few weeks and if we find that ota-monitor is not doing anything other than just removing /hold then we add /hold cancel.

PR #425, opened 2020-09-08
Add suggestions

sdodson:
This adds minor_min, minor_block_list, minor_max, z_min, z_block_list,
and z_max which define the minimum, a list of blocked versions, and a
maximum version which can be used to filter upgrade edges. It's expected
that this is applied atop a list of versions already present in a
channel so the block_list may not see much use.

There's a default as well as arch specific definitions. If you define
arch specific data you are expected to define the entire structure.

### DIFF
--- a/build-suggestions/4.3.yaml
+++ b/build-suggestions/4.3.yaml
@@ -1,0 +1,18 @@
+# valid archs are amd64, s390x
+# If you specify an arch it must have the full set of values
+---
+default:
+  minor_min: 4.2.21
+  minor_block_list: []
+  minor_max: 4.2.9999
+  z_min: 4.3.0
+  z_block_list: []
+  z_max: 4.3.9999
+# s390x:
+#   minor_min: 4.2.21
+#   minor_max: 4.2.9999
+#   minor_block_list:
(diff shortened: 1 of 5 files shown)


sdodson:
To be used with https://github.com/openshift/cincinnati-graph-data/pull/425
Not sure if we should make these arch specific or not?

jupierce:
Having the files include a default with an optional arch override might be easier to maintain.
```
default:
  minor_floor: 4.2.21
  minor_block_list: []
  z_floor: 4.3.0
  z_block_list: [] 
s390x:
  minor_floor: 4.2.26
```

sdodson:
/wip cancel
@jupierce PTAL

sdodson:
/hold cancel
I'd like to move forward with this provided everyone agrees the fields I've defined here provide everything we'd need. I added the max versions so that we can accommodate 4.6 EUS cycle which may potentially have scenarios where we stop shipping 4.7 for a few months even if we're tagging new builds. Since that exact mechanic is as of yet undefined I wanted to provide this flexibility. I've updated https://github.com/openshift/doozer/pull/277 to also support maxes.

PR #433, opened 2020-09-15
Add rust-based CI tool

vrutkovs:
Create a small Rust app which reuses Cincinnati code to verify PRs.

The following checks are implemented:
* Verify files in `channels` and `blocked-edges` are valid yamls, no 
other files are present.
* Check that all files can be deserialized in Cincinnati objects
* Verify that all mentioned releases are pushed to Quay

This will not be hooked up to CI yet, as it requires openshift/release PR to replace python hack.

TODO:
* [x] Requires https://github.com/openshift/cincinnati/pull/304 to merge changes from Cincinnati fork.

/cc @wking @steveeJ


hack/src/check_releases.rs
```diff
@@ -0,0 +1,40 @@
...
+  println!("Verifying all releases are uploaded");
+  for v in found_versions.iter() {
+    if releases
+      .iter()
+      .find(|&r| r.metadata.version == *v)
+      .is_none()
+    {
+      bail!("Version {} is not found in scraped images", v)
+    }
+  }
```
steveeJ, line 37: I think we can be more efficient here by comparing two sets. (Typing in the github comment window here without actually verifying the code, please beware :wink:)

```suggestion
  .context("failed to fetch all release metadata")?
  .into_iter()
  .map(|r| r.metadata.version)
  .collect::<HashSet<Version>>();

  println!("Verifying all releases are uploaded");
  // assume we've assigned the above HashSet to `released_versions`
  let missing_versions = found_versions - released_versions;
  if !missing_versions.is_empty() {
    bail!("Missing the following versions in scraped images: {:?}", missing_versions)
  }
```

hack/src/verify_yaml.rs
```diff
@@ -0,0 +1,127 @@
...
+  println!("Verifying blocked edge files are valid");
+  let blocked_edge_path = data_dir.join(plugin::BLOCKED_EDGES_DIR).canonicalize()?;
+  let blocked_edge_vec =
+    walk_files::<plugin::graph_data_model::BlockedEdge>(&blocked_edge_path, &extension_re).await?;
+  for v in blocked_edge_vec.iter() {
+    found_versions.insert(v.to.clone());
+  }
+
+  println!("Verifying channel files are valid");
+  let channel_path = data_dir.join(plugin::CHANNELS_DIR).canonicalize().unwrap();
```
steveeJ, line 28:
```suggestion
  let channel_path = data_dir.join(plugin::CHANNELS_DIR).canonicalize()?;
```

hack/src/verify_yaml.rs
```diff
@@ -0,0 +1,127 @@
...
+use std::collections::HashSet;
+use std::path::PathBuf;
+use tokio;
+
+pub use anyhow::Result as Fallible;
+use anyhow::{bail, Context};
+
+pub async fn run() -> Fallible<HashSet<Version>> {
+  let data_dir = PathBuf::from("..");
+  let extension_re = Regex::new("ya+ml").unwrap();
```
steveeJ, line 15:
```suggestion
  let extension_re = Regex::new("ya+ml")?;
```

steveeJ (commented):
Thanks for this work! What do you think of creating a sub-directory for the crate separate from _hack_?

hack/src/verify_yaml.rs
```diff
@@ -0,0 +1,127 @@
...
+    ("serialization", serialize_err_vec),
+  ] {
+    if v.1.len() > 0 {
+      println!("Found {} errors:", v.0);
+      println!("{:?}", v.1);
+      has_errors = true;
+    }
+  }
+  match has_errors {
+    true => bail!("Exiting due to errors"),
```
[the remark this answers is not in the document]
vrutkovs: Its probably easier to construct error message here, wdyt?

vrutkovs:
>What do you think of creating a sub-directory for the crate separate from hack

I'm not sure what's the Rust standard for this, but yeah, lets do so

hack/src/main.rs
```diff
@@ -0,0 +1,20 @@
+pub mod check_releases;
+pub mod verify_yaml;
+pub use anyhow::Result as Fallible;
```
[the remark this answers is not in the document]
vrutkovs: Removed extra `pub`s

hack/src/check_releases.rs
vrutkovs: Good idea, implemented this using `HashSet::Difference`

hack/src/verify_yaml.rs
```diff
@@ -0,0 +1,117 @@
...
+    for v in c.versions.iter() {
+      found_versions.insert(v.clone());
+    }
+  }
+
+  Ok(found_versions)
+}
+
+#[derive(Debug)]
+enum ProcessingError {
```
steveeJ, line 41: I think that we don't actually need the `ProcessingError` enum if we just coerce everything to a string at the end. It would in fact be enough if you just appended the error messages to a mutable string throughout the processing. I don't feel strongly about it though so I'm not opposed to keeping this either.

steveeJ (commented):
Are you still deciding whether you move to the crate into its own directory or have you decided not to?

hack/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "cincinnati-graph-data"
+version = "0.1.0"
+authors = ["Vadim Rutkovsky <vrutkovs@redhat.com>"]
+edition = "2018"
```
wking, line 5: Is this some Cargo magic, or should it be 2020?

hack/src/verify_yaml.rs
```diff
@@ -0,0 +1,117 @@
...
+pub async fn run() -> Fallible<HashSet<Version>> {
+  let data_dir = PathBuf::from("..");
+  let extension_re = Regex::new("ya+ml")?;
+  // Collect a list of mentioned versions
+  let mut found_versions: HashSet<Version> = HashSet::new();
+
+  println!("Verifying blocked edge files are valid");
+  let blocked_edge_path = data_dir.join(plugin::BLOCKED_EDGES_DIR).canonicalize()?;
+  let blocked_edge_vec =
+    walk_files::<plugin::graph_data_model::BlockedEdge>(&blocked_edge_path, &extension_re).await?;
```
wking, line 22: Do we need local walkers for validation?  I was expecting us to be able to wash through [the `github-secondary-metadata-scrape` plugin][1].

[1]: https://github.com/openshift/cincinnati/tree/3f5ea5804d1fe67f657e900d652c75622480f9ab/cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper

wking: Somewhat in this space, and feel free to punt as out of scope for a v1 PR, but it would be nice to be able to feed this tool graph-builder and policy-engine plugin configs, which we could copy/edit/paste from our production update service deployment, and have it ingest the local graph-data checkout, wash through the rest of the plugin stack, and spit out assorted policy-engine output graphs for all architectures and channels.  Or diff vs. the current production graph.  Or something...

hack/src/verify_yaml.rs
vrutkovs: I think it would be useful. We might want to add more error types later - i.e. signature check - with customized Display

hack/src/verify_yaml.rs
vrutkovs: metadata-scrape is too forgiving, it intentionally skips files with no extension etc. Perhaps later we could introduce "strict" mode there and merge the walkers.

>it would be nice to be able to feed this tool graph-builder and policy-engine plugin configs

That would be useful to setup configuration (if the registry or org name changes). Probably out of scope now

> spit out assorted policy-engine output graphs for all architectures and channels

that could be done, but its not clear what's the expected graph is. Every change would produce a diff, but its not trivial to validate it. Its out of scope for v1, but might become useful

hack/Cargo.toml
vrutkovs: Rust has two major editions - [2015 and 2018](https://doc.rust-lang.org/edition-guide/editions/index.html)

hack/src/verify_yaml.rs
steveeJ: Adding more types doesn't make it more useful if we still only use it for string coercion :wink: What would convince me if we pass back the error vector to the caller, which would allow it to use the type system to figure out error conditions.

hack/src/verify_yaml.rs
steveeJ: @wking I would've gone in a similar direction to yours but I never got around to pursuing the implementation. My idea was to make a CLI tool very similar to what the GB and PE are today, but without the looping and serving aspects. It would just run the plugin chain once and yield the graph output. The latter could be handled via a plugin as. It might actually be pretty easy to add a config option such as `runonce` to the graphbuilder and add a small plugin which emits the graph to disk, stdout, or whatever we desire.

hack/src/verify_yaml.rs
vrutkovs: >Adding more types doesn't make it more useful 

The idea is to leave formatting errors to Debug impl of ProcessingError. Other tests may extend that struct later.

> What would convince me if we pass back the error vector to the caller

`verify_yamls::run` may throw generic errors, e.g. directory not readable or "blocked-edges" dir cannot be canonicalized.

hack/src/verify_yaml.rs
steveeJ: @wking I came up with https://github.com/openshift/cincinnati/pull/307 which allows the removal of the walker and error types in this PR. @vrutkovs and I decided that this is enough for this PR and would like to postpone further optimization. Is that fine with you?

wking: I am fine punting anything.

hack/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM registry.access.redhat.com/devtools/rust-toolset-rhel7:1.43.1 as builder
+
+WORKDIR /opt/app-root/src/
+COPY . .
+RUN bash -c "source /opt/app-root/etc/scl_enable && cargo build --release"
```
steveeJ, line 5: I don't understand when this switches to the `hack/` directory

steveeJ (changes requested):
Given that we'll further rework this to reuse cincinnati plugin system I think that we're almost good to merge this initial version!

Only two things that I think need changing before we can merge:
 
* It looks like the rust code hasn't run through rustfmt.
* The rust crate is lumped together with the other files in the _hack_ directory.

Could we run the python tool and the rust tool side-by-side for a while to see if they behave the same?

hack/Dockerfile
vrutkovs: This runs within `hack/` directory:
```
- context_dir: hack
  dockerfile_path: Dockerfile
  from: rust-toolset
  to: publish
```
in `release` repo

vrutkovs:
Pushed a fix which formats the code using rustfmt (I'll add a new test to verify that in `release` repo)

steveeJ:
> We'll have to live with that for some time until CI is switched to Rust tool only.

I don't understand why that is. It seems like we're in control over all systems here so we can adjust anything as needed. Would you mind to elaborate?

vrutkovs:
After we merge this tool other PRs might be created, so we can't remove python scripts right away. Also, when `release` PR is created some errors might pop up, so some fixed to this repo might be required. 

Hopefully it would take a day or two to remove all python scripts, so I wouldn't worry about current state too much

steveeJ:
> After we merge this tool other PRs might be created, so we can't remove python scripts right away. Also, when `release` PR is created some errors might pop up, so some fixed to this repo might be required.

I didn't mean to remove the _hack_ directory altogether, just to put the Rust crate in a separate, possibly nested, directory. Wouldn't that work?

vrutkovs:
Why would we want to move it out of `hack`? Eventually it would contain rust pkg only.

I don't mind creating a new dir for that - do you have any catchy name for it in mind?

steveeJ:
> I don't mind creating a new dir for that - do you have any catchy name for it in mind?

Is `graph-data.rs` catchy enough? :smile:

graph-data.rs/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM registry.access.redhat.com/devtools/rust-toolset-rhel7:1.43.1 as builder
 
 WORKDIR /opt/app-root/src/
 COPY . .
-RUN bash -c "source /opt/app-root/etc/scl_enable && cargo build --release"
+RUN bash -c "source /opt/app-root/etc/scl_enable && cargo install --path"
```
steveeJ, line 5: From the help output there seems to be missing a `PATH`

```
        --path <PATH>                                Filesystem path to local crate to install
```

graph-data.rs/src/main.rs
```diff
@@ -0,0 +1,14 @@
...
+use anyhow::Result as Fallible;
+
+async fn run_all_tests() -> Fallible<()> {
+    let found_versions = verify_yaml::run().await?;
+    check_releases::run(&found_versions).await?;
+    Ok(())
+}
+
+fn main() -> Fallible<()> {
+    let mut runtime = tokio::runtime::Runtime::new().unwrap();
```
steveeJ, line 12:
```suggestion
    let mut runtime = tokio::runtime::Runtime::new()?;
```

steveeJ (commented):
This LGTM now! It would be nice to see the CI job in action. What do you think of hooking it up before we merge this PR and marking it as optional so that it doesn't block other PRs?

vrutkovs:
it can't be hooked up before we merge - rehearsals will fail. Lets merge this so that I could land a preliminary version of this and make it "always_run: true" but "optional:true"?

steveeJ:
> it can't be hooked up before we merge - rehearsals will fail. Lets merge this so that I could land a preliminary version of this and make it "always_run: true" but "optional:true"?

/lgtm

steveeJ:
Giving you the opportunity to squash:

/hold

steveeJ (approved):
/lgtm

PR #434, opened 2020-09-15
hack/errata.py: trap exceptions in lgtm_fast_pr_for_errata

vrutkovs:
Errata-hook deployment on datahub is crashing and not notifying us correctly. This ensures exception is trapped, more debug messages added, and PRs which already have lgtm label are skipped.

### DIFF
--- a/hack/errata.py
+++ b/hack/errata.py
@@ -44,7 +44,10 @@ def run(poll_period=datetime.timedelta(seconds=3600),
         for message in poll(period=2*poll_period, **kwargs):
             if cache and message['fulladvisory'] in cache or 'bug fix update' not in message['synopsis']:
                 continue
-            message['approved_pr'] = lgtm_fast_pr_for_errata(githubrepo, githubtoken, message)
+            try:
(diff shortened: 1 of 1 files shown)


hack/errata.py
```diff
@@ -44,7 +44,10 @@ def run(poll_period=datetime.timedelta(seconds=3600),
         for message in poll(period=2*poll_period, **kwargs):
             if cache and message['fulladvisory'] in cache or 'bug fix update' not in message['synopsis']:
                 continue
-            message['approved_pr'] = lgtm_fast_pr_for_errata(githubrepo, githubtoken, message)
+            try:
+                message['approved_pr'] = lgtm_fast_pr_for_errata(githubrepo, githubtoken, message)
+            except Exception as error:
+                _LOGGER.error('Error looking up PRs: {}'.format(error))
```
wking, line 50: nit: As a non-fatal complaint that might be due to handling some unrelated PR, this should be `_LOGGER.warn`.

vrutkovs: Updated, thanks

wking (approved):
/lgtm

PR #450, opened 2020-09-22
hack/errata: strip newlines in argparse arguments

vrutkovs:
erratahoook deployment on datahub adds extra `\n` to the github token stored in the secret in the resulting queries. I checked the secret and it doesn't contain extra new line, so it must be python argparse doing that.

This would ensure received arguments are stripped.

/cc @wking

### DIFF
--- a/hack/errata.py
+++ b/hack/errata.py
@@ -204,9 +204,9 @@ if __name__ == '__main__':
     try:
         run(
             cache=cache,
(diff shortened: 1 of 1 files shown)


wking (approved):
I would be very surprised if argparse is adding newlines, but stripping here is a safe enough backstop for wherever they're being added.

/lgtm

PR #484, opened 2020-10-02
Blocking edges from 4.4.* to 4.5.*

jottofar:


LalatenduMohanty (changes requested):
We should block for all nodes present in https://github.com/openshift/cincinnati-graph-data/blob/master/channels/fast-4.5.yaml including 4.5.0-0.hotfix-2020-08-24-185832

blocked-edges/4.5.9.yaml
```diff
@@ -1,3 +1,3 @@
 to: 4.5.9
-from: 4\.4\.21
-# Kernel and other packages older in 4.5.9 than 4.4.21, https://bugzilla.redhat.com/show_bug.cgi?id=1879260
```
LalatenduMohanty, line 3: We do not want to remove this comment for historical reasons.

LalatenduMohanty (changes requested):

wking:
Preliminary data suggests [rhbz#1883589][1] is rare and has [a simple mitigation][2], so we're going to sit on these edge-blocks to give the API-server folks some time to work up future plans.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1883589#c15
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1883589#c18

blocked-edges/4.5.11.yaml
```diff
@@ -0,0 +1,3 @@
+to: 4.5.11
+from: 4\.4\.*
```
wking, line 2: you need `4\.4\..*` here, and throughout this PR.  Compare:

```console
$ cat blocked-edges/4.2.19.yaml     
to: 4.2.19
from: 4\.1\..*
# 4.1 -> 4.2 -> 4.3 updates hit TargetDown for ingress-operator, fixed in 4.2.21, https://bugzilla.redhat.com/show_bug.cgi?id=1801300
```

sdodson:
4.5.16 includes some of the proposed fixes for the problems that prompted this potential edge revocation but it also includes upgrade edges from all previous 4.4.z as well, so in theory everyone will start choosing this upgrade path in the future. However if we see that's not the case AND we see that the fixes delivered in 4.5.16 have a meaningful impact on KubeAPIDown alerts we may want to move forward with blocking upgrades to anything older than 4.5.16.

API server team has been asked to review telemetry data for impact by end of week, lets sync up with them and make a final decision to move forward here or close this.

sdodson:
The apiserver team is focused on other items and will not complete the analysis we've asked for. Without them advocating for this block we're going to close this.
/close

PR #487, opened 2020-10-07
graph-data.rs: pin protobuf and protoc

vrutkovs:
Use protoc/protobuf 2.8.0 compatible with Rust 1.43. This fixes rehearsal issue in https://github.com/openshift/release/pull/12491

### DIFF
--- a/graph-data.rs/Cargo.toml
+++ b/graph-data.rs/Cargo.toml
@@ -13,3 +13,5 @@ anyhow = "1.0"
 regex = "^1.1.0"
 semver = { version = "^0.9.0", features = [ "serde" ] }
 toml = "^0.4.10"
+protoc = "~2.8.0"
+protobuf = "~2.8.0"
--- a/graph-data.rs/Cargo.lock
+++ b/graph-data.rs/Cargo.lock
@@ -4,456 +4,535 @@
 name = "actix-codec"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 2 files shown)


graph-data.rs/Cargo.toml
```diff
@@ -13,3 +13,5 @@ anyhow = "1.0"
 regex = "^1.1.0"
 semver = { version = "^0.9.0", features = [ "serde" ] }
 toml = "^0.4.10"
+protoc = "=2.8.0"
```
steveeJ, line 16: I think we could still allow patch release updates

```suggestion
protoc = "^2.8.0"
```

graph-data.rs/Cargo.toml
```diff
@@ -13,3 +13,5 @@ anyhow = "1.0"
...
+protobuf = "=2.8.0"
```
steveeJ, line 17:
```suggestion
protobuf = "^2.8.0"
```

graph-data.rs/Cargo.toml
vrutkovs: Not sure why, but it makes `cargo generate-lockfiles` bring in the cursed 2.18.0. Tilde notation makes it update to `2.8.2`, which is what we want

vrutkovs: Ah, yes, caret notation would fix the major version only - `^1.2.3  :=  >=1.2.3, <2.0.0`, so `^2.8.0 := >=2.8.0, < 3.0.0`, which is undesirable

steveeJ (approved):
/lgtm

PR #490, opened 2020-10-08
Graphdata rework dockerfile

vrutkovs:
Previously Rust CI tool was using `graph-data.rs` context, so that it would produce an image with just the binary. The plan was to promote it to `cincinnati-ci` namespace and avoid rebuilding it to save time.
However that didn't really work well:
* CI can't do optional builds atm, so there is no way to skip the build even if `graph-data.rs` is not touched
* There is no easy way to run the image with other src from the repo. CI provides a way to build `bin` container from `src`, but currently it would break existing python script, so implementation of that needs to be postponed.

This PR would rework graph-data Dockerfile so that it would include the whole repo and adjust the test to run from current directory (previously the binary was expected to run from the subfolder)

/cc @steveeJ

### DIFF
--- a/graph-data.rs/src/verify_yaml.rs
+++ b/graph-data.rs/src/verify_yaml.rs
@@ -7,7 +7,11 @@ use std::path::PathBuf;
 use anyhow::Result as Fallible;
 
 pub async fn run() -> Fallible<HashSet<Version>> {
-    let data_dir = PathBuf::from("..");
+    let data_dir = PathBuf::from(".");
+    println!(
+        "Looking for metadata in {:?}",
+        data_dir.canonicalize()?
+    );
     let all_files_regex = Regex::new(".*")?;
     let disallowed_errors: HashSet<plugin::DeserializeDirectoryFilesErrorDiscriminants> = [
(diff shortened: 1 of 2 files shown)


graph-data.rs/src/verify_yaml.rs
```diff
@@ -7,7 +7,11 @@ use std::path::PathBuf;
 use anyhow::Result as Fallible;
 
 pub async fn run() -> Fallible<HashSet<Version>> {
-    let data_dir = PathBuf::from("..");
+    let data_dir = PathBuf::from(".");
+    println!(
+        "Looking for metadata in {:?}",
+        data_dir.canonicalize().unwrap()
```
steveeJ, line 13: IMO `unwrap` should be a last resort. I think here we can use `?` without an issue.

```suggestion
        data_dir.canonicalize()?
```

steveeJ (changes requested):

steveeJ (approved):
/lgtm

vrutkovs:
/hold cancel

e2e passed, although I don't see the logs

PR #493, opened 2020-10-12
Block edges to older 4.6.0-fc and rc.2 from 4.5.11 and 4.5.13

sdodson:
This will get us as close to the final GA edges as we can be right
now. Once 4.5.15 enters the candidate channel we'll block edges from
4.5.14 as well.

Benefit of this is that we ensure anyone testing on graph from this point
forward provides us the best signal for GA readiness possible.


sdodson:
/hold
Need to ensure that this doesn't block any of OSD's testing efforts.

sdodson:
/close
We'll just focus on getting edges for GA right like in #521

PR #512, opened 2020-10-21
graph-data.rs/Dockerfile: run cargo install as root

vrutkovs:
In newer openshift versions builds must explicitly declare a user,
otherwise these would run as unprivileged user. However copied source
would be owned by root, so `target/release` cannot be created.


LalatenduMohanty (approved):
/lgtm

PR #527, opened 2020-10-30
graph-data.rs: ensure signatures are present for all scraped images

vrutkovs:
This uses several constants from CVO to verify that all images have a
signature file present. This doesn't verify the that the signature is
valid though.

TODO:
* [x] Ask ART to [have a look](https://issues.redhat.com/browse/ART-2397) at current failures - these seem valid and need fixing

Ref: https://issues.redhat.com/browse/OTA-153

### DIFF
--- a/graph-data.rs/src/check_signatures.rs
+++ b/graph-data.rs/src/check_signatures.rs
@@ -19,10 +19,25 @@ lazy_static! {
       .expect("could not parse url");
 }
 
+// Signature file request timeout
 static DEFAULT_TIMEOUT_SECS: u64 = 30;
 
 // CVO has maxSignatureSearch = 10 in pkg/verify/verify.go
 static MAX_SIGNATURES: u64 = 10;
+
+// Skip some versions from 4.0 / 4.1 / 4.2 times
+// https://issues.redhat.com/browse/ART-2397
+static SKIP_VERSIONS: &[&str] = &[
+  "4.1.0-rc.3+amd64",
+  "4.1.0-rc.5+amd64",
+  "4.1.0-rc.4+amd64",
+  "4.1.0-rc.0+amd64",
+  "4.1.0-rc.8+amd64",
+  "4.1.37+amd64",
+  "4.2.11+amd64",
+  "4.3.0-rc.0+amd64",
+  "4.6.0-fc.3+s390x",
+];
 
 fn payload_from_release(release: &Release) -> Fallible<String> {
   match release {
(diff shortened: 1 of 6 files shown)


graph-data.rs/src/check_signatures.rs
```diff
@@ -0,0 +1,127 @@
...
+fn payload_from_release(release: &Release) -> Fallible<String> {
+  match release {
+    Release::Concrete(c) => Ok(c.payload.clone()),
+    _ => Err(format_err!("not a concrete release")),
+  }
+}
+
+async fn fetch_url(client: &Client, sha: &str, i: u64) -> Fallible<()> {
+  let url = BASE_URL
+    .join(format!("sha256={}/", sha).as_str())?
```
wking, line 36: nit: this assumes that we will stick with sha256.  Safe enough in the near term, but it would be more robust to have this take the full digest and [replace the `:` with `=`][1].

[1]: https://github.com/containers/image/blame/f1a6f9b50d0286bf208935d4b9ab48117294ac31/docs/signature-protocols.md#L39-L51

graph-data.rs/src/check_signatures.rs
```diff
@@ -0,0 +1,149 @@
...
+
+// CVO has maxSignatureSearch = 10 in pkg/verify/verify.go
+static MAX_SIGNATURES: u64 = 10;
+
+// Skip some versions from 4.0 / 4.1 / 4.2 times
+// https://issues.redhat.com/browse/ART-2397
+static SKIP_VERSIONS: &[&str] = &[
+  "4.0.0-0.9+amd64",
+  "4.0.0-0.10+amd64",
+  "4.0.0-0.11+amd64",
```
wking, line 33: no need to skip these since #528 , right?

graph-data.rs/src/check_signatures.rs
```diff
@@ -0,0 +1,149 @@
...
+  "4.1.0-rc.3+amd64",
+  "4.1.0-rc.5+amd64",
+  "4.1.0-rc.4+amd64",
+  "4.1.0-rc.0+amd64",
+  "4.1.0-rc.8+amd64",
+  "4.2.11+amd64",
+  "4.3.0-rc.0+amd64",
+  "4.1.37+amd64",
```
wking, line 41: nit: the rest of your skip entries seem to be mostly ordered by increasing version.  Can we shift this before the 4.2.11 entry?

vrutkovs: Done

graph-data.rs/src/check_signatures.rs
vrutkovs: Ah, correct. Removed those, lets see

wking (approved):
/lgtm

vrutkovs:
/hold

Waiting for e2e to pass

PR #541, opened 2020-11-09
graph-data.rs: verify signature when checking release

vrutkovs:
This checks that release signature can be verified using release public keys. `add_public_keys_to_keyring` adds release keys to gpg keyring and `verify_signature` ensures the signature is correct. This function also checks that the message digest is correct.

Added `4.1.0+amd64` to `SKIP_VERSIONS` list as its apparently signed with CI key

Ref: https://issues.redhat.com/browse/OTA-153


wking:
[e2e][1] has lots of:

```
    Err(
        Error {
            context: "Verifying signature",
            source: Error {
                context: "parsing after at around 1",
                source: "not supported: new format",
            },
        },
    ),
```

That's `gpgrv` choking on the payloads?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-graph-data/541/pull-ci-openshift-cincinnati-graph-data-master-e2e/1325796425899970560#1:build-log.txt%3A26

vrutkovs:
Yup, seems we're using new gpg format, not yet implemented there.

I'll give `rPGP` a try

vrutkovs:
Seems `4.1.0+amd64` is signed with CI key - however I don't know why is it still showing up in our channels :/

graph-data.rs/src/check_signatures.rs
```diff
@@ -39,32 +43,39 @@ static SKIP_VERSIONS: &[&str] = &[
   "4.6.0-fc.3+s390x",
 ];
 
+/// Extract payload value from Release if its a Concrete release
```
wking, line 48: Nit: "its" -> "it is"

graph-data.rs/src/gpg.rs
```diff
@@ -0,0 +1,85 @@
...
+pub async fn verify_signature(
+  public_keys: &Keyring,
+  body: Bytes,
+  expected_digest: &str,
+) -> Fallible<()> {
+  let msg = Message::from_bytes(body.reader()).context("Parsing message")?;
+
+  // Verify signature using provided public keys
+  if !public_keys.iter().any(|ref k| msg.verify(k).is_ok()) {
+    return Err(format_err!("No matching key found to decrypt {:#?}", msg));
```
wking, line 67: s/decrypt/verify/?

Also, we can extract the claimed signer's key ID [from the message][1], so we shouldn't need to iterate over trusted keys; we can just verify vs. that claimed key (or complain that the claimed key ID is unrecognized).  But iterating over a handful of keys should be cheap enough, if the PGP library doesn't expose the issuer conveniently.

[1]: https://tools.ietf.org/html/rfc4880#section-5.2.3.5

vrutkovs: Ah, good idea, seems we can fetch KeyID via https://docs.rs/pgp/0.7.1/pgp/composed/message/enum.Message.html#method.get_recipients

vrutkovs: Hmm, it returns empty and `key_id` is private is most signature types.

I'd rather leave it for later

wking (approved):
/lgtm

PR #549, opened 2020-11-12
channels: More backfilling (e.g. 4.3.0 in candidate-4.4)

wking:
Generated with:

```console
$ hack/backfill.py
$ git checkout HEAD -- channels/candidate-4.3.yaml
$ emacs channels/candidate-4.4.yaml
```

With the candidate-4.3 and 4.4 massaging so we don't get:

```diff
 - 4.3.14
+- 4.3.14+amd64
```

and similar, where all arches went into candidate-4.3, but only amd64 went into the candidate-4.4, 5ff491a42f (#193).


channels/candidate-4.5.yaml
```diff
@@ -23,7 +23,25 @@ versions:
...
 - 4.4.9
 - 4.4.8
+- 4.4.7
 - 4.4.6
+- 4.4.5
+- 4.4.4
+- 4.4.3
+- 4.4.2
+- 4.4.0
+- 4.4.0-rc.0
```
sdodson, line 33: I thought we were only adding versions from the previous minor if they had made it to stable channel?

wking: I'm only adding versions *to stable* from the previous minor if they were in the previous minor's stable channel.  I'm adding versions to candidate from the previous minor if they made it to the previous minor's candidate channel, etc.  This means that we're basically making one call about stability, and then putting the release in all 4.y and 4.(y+1) channels that match that stability decision.

sdodson: Yeah but we never tested upgrades from 4.4.0-rc.0 to any 4.5 version outside of immediately after branching which is usually a dumpster fire anyway.

sdodson: bah, nevermind, that doesn't mean we'd have edges, guess this is fine though confusing

channels/stable-4.6.yaml
```diff
@@ -1,5 +1,21 @@
 name: stable-4.6
 versions:
+- 4.5.18
```
sdodson, line 3: We need to hold these until we provide upgrades to 4.6. Lets drop these and ensure that our PR to bring upgrades to stable-4.6 includes them?

wking: Fixed with 6408ba1 -> 1d9fe0f.

channels/candidate-4.5.yaml
wking: yeah, hopefully this PR adds no edges.  The idea is just to let folks say "I would like to be on $next_minor" regardless of which release they are on in the current minor, vs. today when they have to update to some threshold patch release before they can change channels.

PR #550, opened 2020-11-16
Block 4.5 to 4.6.1 and 4.6.3

sdodson:
4.5.16+ to 4.6.4 is the best known upgrade path at this point in
time

### DIFF
--- a/blocked-edges/4.5.14-15-to-4.6.1.yaml
+++ b/blocked-edges/4.5.14-15-to-4.6.1.yaml
@@ -1,3 +1,0 @@
-to: 4.6.1
-from: 4\.5\.1[45]
-# Need 4.5.16+ to upgrade to 4.6 for Create a drop-in file for cri-o’s default_capabilities https://bugzilla.redhat.com/show_bug.cgi?id=1856529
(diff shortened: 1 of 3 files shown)


blocked-edges/4.5.14-15-to-4.6.1.yaml
```diff
@@ -1,3 +0,0 @@
-to: 4.6.1
-from: 4\.5\.1[45]
-# Need 4.5.16+ to upgrade to 4.6 for Create a drop-in file for cri-o’s default_capabilities https://bugzilla.redhat.com/show_bug.cgi?id=1856529
```
wking, line 3: Can we preserve this comment in `blocked-edges/4.5-to-4.6.1.yaml`?  So if some future devs decide 1893803 is not block-worthy, they don't forget that there were other reasons we blocked some 4.5 -> 4.6.1.

Also, because these blocks will always (I think) be consolidating all reasons for blocking edges into a given release, I'd rather just name them `{target-version}.yaml`, instead of `{sources}-to-{target-version}.yaml`.  Thoughts?

sdodson: Yeah that's fine with me.

wking (approved):
/lgtm

PR #557, opened 2020-11-19
Add eus-4.6 channel

sdodson:
This channel will mirror the stable-4.6 channel until such a time
as 4.6 transitions to EUS status.


channels/eus-4.6.yaml
```diff
@@ -0,0 +1,5 @@
+name: eus-4.6
```
wking, line 1: `extended-update-support-4.6`?  Or are there benefits to squeezing this down to fewer characters?

wking:
Why the WIP?  I am personally fine landing this whenever.  Would be nice if ART's stablization PRs would fill both stable-4.6 and this channel, although manually syncing for a while wouldn't be the end of the world.  I think folks also had some ideas about when to provide this channel (once 4.6 enters maintenance mode?), but I don't remember why that delay was useful, vs. just creating the channel now.

channels/eus-4.6.yaml
sdodson: I'm not opposed to that but "eus-4.6" is what I'd proposed and everyone has agreed so far. I'm sure such a long string would wrap in the current console UX and I see this more as a channel id than a full description. We should get those descriptive channel text api in :-)

sdodson:
There's a card for ART to handle that, I need to get all of this approved by multiple parties and documentation changes ready before we land this.

Epic is here https://issues.redhat.com/browse/OTA-325

LalatenduMohanty:
@sdodson Should we wait for 4.6.8, to have an edge to stable first?

sdodson:
> @sdodson Should we wait for 4.6.8, to have an edge to stable first?

yes, this will be delayed because 4.6.7 was blocked

sdodson:
/hold cancel
4.6.8 has gone out and the errata notes mention addition of this channel so lets land this. I'll update the 4.6.8 stable channel PR to include updating eus-4.6 channel so that when it goes out it does the same.

PR #513, opened 2020-10-21
hack/errata: include erratas with security update

vrutkovs:
Some erratas may include security fixes, so their format changes from
```
OpenShift Container Platform 4.3.40 bug fix update
```
to
```
Low: OpenShift Container Platform 4.3.40 security and bug fix update
```

This PR optionally matches impact and "security and " stanza.

PR #558, opened 2020-11-24
hack/errata: Match '... bug fix and golang security update' synopsis

wking:
To match:

    Moderate: OpenShift Container Platform 4.5.20 bug fix and golang security update

while continuing to exclude:

    OpenShift Container Platform 4.5.11 optional CSI driver Operators bug fix update

from ee59956db5 (#480).  I'm not sure why the 4.5.20 approach differs from:

    Low: OpenShift Container Platform 4.3.40 security and bug fix update

from d161733812 (#513), but we don't get to name these, we just have to match the names other folks pick.

Also adds a commit to log excluded synopses, to make debugging this sort of match failure easier going forward.


vrutkovs:
/lgtm

Perhaps we should switch to denylists (i.e. "optional" in title) instead of matching the text which is highly volatile?

PR #576, opened 2020-12-03
Block 4.5 to 4.6.4

sdodson:
We fixed ~ 35 bugs in 4.6.6 and one of them was a prometheus storage
utilization regression from 4.5, so lets funnel everyone to the best
4.6.


sdodson:
Had some special characters in the comment, removed them no changes to actual content so reapplying lgtm

PR #577, opened 2020-12-04
blocked-edges/4.6.6: Block * -> 4.6.(6|7) on born-in-4.1 ingress

wking:
There aren't all that many born-in-4.1 clusters within striking distance of 4.6.6+, so the number of vulnerable clusters is small. But `*.apps` breaking would be really bad, so block the edges to protect that small subset.  [rhbz#1904582](https://bugzilla.redhat.com/show_bug.cgi?id=1904582).

We could keep 4.6.6 -> 4.6.7 in place, but I didn't bother to figure out the regexp for "all versions that aren't 4.6.6".

### DIFF
--- a/blocked-edges/4.6.6.yaml
+++ b/blocked-edges/4.6.6.yaml
@@ -1,0 +1,3 @@
+to: 4.6.6
+from: .*
+# *.apps ingress breaks for born-in-4.1 clusters that incorrectly transition to internal DNS on 4.6.6 and later, https://bugzilla.redhat.com/show_bug.cgi?id=1904582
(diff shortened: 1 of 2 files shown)


wking:
[publish][1] `Name or service not known`.  Dunno what that's about.

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-graph-data/577/pull-ci-openshift-cincinnati-graph-data-master-publish/1334967105304399872#1:build-log.txt%3A158

wking:
e2e passed, I don't care enough about publish to dig in.

/override ci/prow/publish

PR #626, opened 2021-01-21
graph-data.rs: verify channel order

vrutkovs:
This adds a CI tests which ensures, that releases are properly ordered: first it must be placed in candidate, then fast and then stable. If a release is put in candidate and then in stable, it would throw an error.

EUS channel is omitted for now, as its not yet entirely clear how releases would be structured there.

https://issues.redhat.com/browse/OTA-126

/cc @LalatenduMohanty @wking @PratikMahajan


graph-data.rs/src/check_channels.rs
```diff
@@ -0,0 +1,53 @@
...
+    for r in releases {
+        let release_vec: Vec<String> = CHANNEL_ORDER
+            .iter()
+            .map(|v| format!("{}-{}", v, r))
+            .collect();
+        // Iterate over pairs
+        for (a, b) in release_vec.iter().tuple_windows() {
+            // Check that channels are present
+            if channels.get(a).is_none() || channels.get(b).is_none() {
+                continue;
```
wking, line 37: If `a` is present, it is an error for `b` to be absent (e.g. non-empty stable-4.6 requires fast 4.6 to exist).

graph-data.rs/src/check_channels.rs
```diff
@@ -0,0 +1,53 @@
...
+    // Collect a list of releases (4.6, 4.5 etc.)
+    let releases: Vec<&str> = channels_vec
+        .iter()
+        .filter_map(|c| c.name.split("-").last())
+        .unique()
+        .collect();
+
+    // Prepare a list of expected channels, ordered according to the rule
+    for r in releases {
+        let release_vec: Vec<String> = CHANNEL_ORDER
```
wking, line 29: Special-case 4.1, where we had `presubmit-4.1` and `stable-4.1`?

graph-data.rs/src/check_channels.rs
vrutkovs: This check makes sense, but it currently fails: `Channel stable-4.1 exists, but not fast-4.1`.

I wonder if I should :
* add `fast-4.1` and `candidate-4.1`
* treat "prerelease" as "fast and candidate"
* or ignore 4.1 all together

(picking the latter for now)

graph-data.rs/src/check_channels.rs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            errors.push(e.unwrap())
+        }
+    }
+    if errors.is_empty() {
+        Ok(())
+    } else {
+        Err(anyhow::anyhow!("Error in channel ordering: {:?}", errors))
+    }
+}
```
LalatenduMohanty, line 75: @vrutkovs Now that we have some sizable code here we should add unit tests. It does not have to go in this PR but we should add unit tests soon.

graph-data.rs/src/check_channels.rs
LalatenduMohanty: I guess so. @vrutkovs can confirm.

graph-data.rs/src/check_channels.rs
vrutkovs: Good idea, filed OTA-342

LalatenduMohanty (approved):
/lgtm

PR #647, opened 2021-02-03
build-suggestions: Update 4.7

sdodson:
This will likely be updated again before GA, at least z_min.

### DIFF
--- a/build-suggestions/4.7.yaml
+++ b/build-suggestions/4.7.yaml
@@ -2,10 +2,10 @@
 # If you specify an arch it must have the full set of values
 ---
 default:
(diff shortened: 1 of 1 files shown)


sdodson:
@wking @jottofar Lets go ahead and tighten these up now that [ART is actually using these suggestions](https://github.com/openshift/doozer/pull/277). No specific reason for 4.6.15 other than it's relatively recent, so we've shaken out most of the bad 4.6 bugs, and it's currently the widest usage. If we're aware of any definitive reasons to raise the minimums lets get those detailed somewhere.

LalatenduMohanty (approved):
/lgtm

PR #668, opened 2021-02-24
Enable 4.7.0 in stable channel(s)

wking:
No need to cook in fast, because there will be no stable-4.7 into or out of 4.7.0 for a while.

### DIFF
--- a/channels/stable-4.7.yaml
+++ b/channels/stable-4.7.yaml
@@ -1,0 +1,3 @@
+name: stable-4.7
+versions:
+- 4.7.0


jeremyeder:
The OSD and ROSA products only support the stable channel.  In order for those products to pick up 4.7, an entry has to be in the stable channel to support new installs.  During the 4.6 GA, this is what happened.  An entry showed up in both fast and stable at the time of GA.  We are not talking about upgrades at all here.

wking:
Ah, this is a dup of #663

/close

PR #676, opened 2021-03-04
errata-webhook: add initial unit tests

Davoska:
Initial unit tests for the errata-webhook.

Pull request related to https://issues.redhat.com/browse/OTA-356.

### DIFF
--- a/hack/requirements.txt
+++ b/hack/requirements.txt
@@ -1,1 +1,2 @@
 PyGithub==1.52
+pytest==6.2.2
--- a/hack/requirements.txt
+++ b/hack/requirements.txt
@@ -1,2 +1,1 @@
 PyGithub==1.52
-pytest==6.2.2
--- a/hack/test_errata.py
+++ b/hack/test_errata.py
@@ -91,9 +91,9 @@ def test_poll_params_of_url(urlopen_mock, json_load_mock):
     parsed_url = urllib.parse.urlparse(url)
     params = urllib.parse.parse_qs(parsed_url.query)
 
-    # Assert if parametres complies with datagrepper reference
+    # Assert if parameters complies with datagrepper reference
     assert int(params["page"][0]) > 0               # Page must be greater than 0
-    assert int(params["rows_per_page"][0]) <= 100   # Must be les than or equal to 100
+    assert int(params["rows_per_page"][0]) <= 100   # Must be less than or equal to 100
     assert params["category"][0] == "errata"        # Should only look for errata category
     assert params["contains"][0] == "RHOSE"         # Only messages containing RHOSE
 
@@ -117,7 +117,7 @@ def test_poll_params_of_url(urlopen_mock, json_load_mock):
 #         "pages": 1
 #     }
     
-#     # Retrive messages from errata.poll without specifying parametres
+#     # Retrive messages from errata.poll without specifying parameters
 #     polled_messages = []
 #     for message in errata.poll():
 #         polled_messages.append(message)
@@ -208,7 +208,7 @@ def test_poll_multiple_messages(urlopen_mock, json_load_mock):
     urlopen_mock.return_value = MagicMock()
     raw_messages = [
         {
-            "additional_necessary_info": "shouldn't be proccesed",
+            "additional_necessary_info": "shouldn't be processed",
             "msg": {
                 "errata_id": 1,
                 "product": "RHOSE",
@@ -216,7 +216,7 @@ def test_poll_multiple_messages(urlopen_mock, json_load_mock):
             }
         },
         {
-            "additional_necessary_info": "shouldn't be proccesed",
+            "additional_necessary_info": "shouldn't be processed",
             "msg": {
                 "errata_id": 2,
                 "product": "RHOSE",
@@ -224,7 +224,7 @@ def test_poll_multiple_messages(urlopen_mock, json_load_mock):
             }
         },
         {
-            "additional_necessary_info": "shouldn't be proccesed",
+            "additional_necessary_info": "shouldn't be processed",
             "msg": {
                 "errata_id": 3,
                 "product": "RHEL",
@@ -232,7 +232,7 @@ def test_poll_multiple_messages(urlopen_mock, json_load_mock):
             }
         },
         {
-            "additional_necessary_info": "shouldn't be proccesed",
+            "additional_necessary_info": "shouldn't be processed",
             "msg": {
                 "errata_id": 4,
                 "product": "RHEL",
@@ -240,7 +240,7 @@ def test_poll_multiple_messages(urlopen_mock, json_load_mock):
             }
         },
         {
-            "additional_necessary_info": "shouldn't be proccesed",
+            "additional_necessary_info": "shouldn't be processed",
             "msg": {
                 "errata_id": 5,
                 "product": "RHOSE",
--- a/hack/test_errata.py
+++ b/hack/test_errata.py
@@ -98,35 +98,6 @@ def test_poll_params_of_url(urlopen_mock, json_load_mock):
     assert params["contains"][0] == "RHOSE"         # Only messages containing RHOSE
 
 
-# Possible minor bug, calling errata.poll() without passing arguments
-# @patch("json.load")
-# @patch("urllib.request.urlopen")
-# def test_poll_default_params(urlopen_mock, json_load_mock):
-#     urlopen_mock.return_value = MagicMock()
-#     # Mocked response from API
-#     raw_messages = [
-#         {
-#             "msg": {
-#                 "product": "RHOSE",
-#                 "to": "SHIPPED_LIVE",
-#             }
-#         }
-#     ]
-#     json_load_mock.return_value = {
-#         "raw_messages": raw_messages,
-#         "pages": 1
-#     }
-    
-#     # Retrive messages from errata.poll without specifying parameters
-#     polled_messages = []
-#     for message in errata.poll():
-#         polled_messages.append(message)
-
-#     # Test if data was polled correctly
-#     expected_messages = [x['msg'] for x in raw_messages]
-#     assert expected_messages == polled_messages
-
-
 @patch("json.load")
 @patch("urllib.request.urlopen")
 def test_poll_number_of_returned_pages_is_zero(urlopen_mock, json_load_mock):
--- a/hack/test_errata.py
+++ b/hack/test_errata.py
@@ -1,0 +1,598 @@
+import datetime
+import errata
+import github
+import http
+import json
+import os
+import re
+import tempfile
+import unittest
+import urllib
+from collections import namedtuple
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+def test_extract_errata_number_from_body_valid_url():
+    body = "errata is https://errata.devel.redhat.com/advisory/12345"
+    assert errata.extract_errata_number_from_body(body) == 12345
+
+    body = "errata is https://errata.devel.redhat.com/advisory/9999"
+    assert errata.extract_errata_number_from_body(body) == 9999
+
+    body = "errata is https://errata.devel.redhat.com/advisory/"
+    assert errata.extract_errata_number_from_body(body) == None
+
+    body = "errata is https://errata.devel.redhat.com/advisory/invalid"
+    assert errata.extract_errata_number_from_body(body) == None
+
+
+def test_extract_errata_number_from_body_invalid_url():
+    # Possible invalid body
+    # body = "errata is https://errata.devel.redhat.com/advisory/advisory/12345"
+    # assert errata.extract_errata_number_from_body(body) == None
+
+    body = "errata is https://errata.devel.redhat.com/"
+    assert errata.extract_errata_number_from_body(body) == None
+
+
+def test_load():
+    with tempfile.TemporaryDirectory() as tempdir:
+        cachepath = os.path.join(tempdir, "cache.json")
+        cache = {"foo": "bar"}
+
+        with open(cachepath, 'w') as file:
+            json.dump(cache, file)
+
+        loaded_cache = errata.load(cachepath)
+        assert loaded_cache == cache
+
+
+def test_load_nonexisting_file():
+    with tempfile.TemporaryDirectory() as tempdir:
+        return_val = errata.load(os.path.join(
+            tempdir, "non_existing_file.txt"))
+        assert return_val == {}
+
+
+def test_save():
+    with tempfile.TemporaryDirectory() as tempdir:
+        cache = {"foo": "bar"}
+        cachepath = os.path.join(tempdir, "cache.json")
+        errata.save(cachepath, cache)
+        assert os.path.isfile(cachepath)
+
+        with open(cachepath, "r") as file:
+            loaded_cache = json.load(file)
+            assert cache == loaded_cache
+
+
+@patch("json.load")
+@patch("urllib.request.urlopen")
+def test_poll_params_of_url(urlopen_mock, json_load_mock):
+    urlopen_mock.return_value = MagicMock()
+    
+    # Mocked response from API
+    raw_messages = []
+    json_load_mock.return_value = {
+        "raw_messages": raw_messages,
+        "pages": 1
+    }
+    
+    # Call errata.poll function
+    polled_messages = []
+    poll_period = datetime.timedelta(seconds=3600)
+    for message in errata.poll(period=poll_period):
+        polled_messages.append(message)
+
+    #Get arguments of urlib.request.urlopen call inside errata.poll
+    urlopen_called_with_args = urlopen_mock.call_args
+    # Get params of the url
+    url = re.search(r"call\(\'(.*)\'\)", str(urlopen_called_with_args)).group(1)
+    parsed_url = urllib.parse.urlparse(url)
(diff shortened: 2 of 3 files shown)


hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+
+        with open(cachepath, 'w') as file:
+            json.dump(cache, file)
+
+        loaded_cache = errata.load(cachepath)
+        assert loaded_cache == cache
+
+
+def test_load_nonexisting_file():
+    with tempfile.TemporaryDirectory() as tempdir:
```
vrutkovs, line 51: [tmpdir fixture](https://docs.pytest.org/en/stable/tmpdir.html)

hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+    params = urllib.parse.parse_qs(parsed_url.query)
+
+    # Assert if parametres complies with datagrepper reference
+    assert int(params["page"][0]) > 0               # Page must be greater than 0
+    assert int(params["rows_per_page"][0]) <= 100   # Must be les than or equal to 100
+    assert params["category"][0] == "errata"        # Should only look for errata category
+    assert params["contains"][0] == "RHOSE"         # Only messages containing RHOSE
+
+
+# Possible minor bug, calling errata.poll() without passing arguments
```
vrutkovs, line 101: I don't think its critical, errata.poll is not meant to be used externally

hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+        "fulladvisory": "RHEL",
+        "when": "today",
+        "synopsis": "OpenShift Container Platform X.X.X",
+        "uri": "example_uri",
+        "approved_pr": "example_approved"
+    }
+
+    errata.notify(message, True)
+
+    msg_text = '<!subteam^STE7S7ZU2>: {fulladvisory} shipped {when}: {synopsis} {uri}'.format(**message)
```
vrutkovs, line 347: This needs to be a template set in `errata.py` which we'd reuse in unit tests

hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+    repo.get_pulls = MagicMock(return_value=prs)
+
+    get_prs = []
+    for pr in errata.get_open_prs_to_fast(repo):
+        get_prs.append(pr)
+
+    assert get_prs == prs
+
+
+def test_get_open_prs_to_fast_labels_lgtmed():
```
vrutkovs, line 412: Not quite following what does it do - this checks that lgtm'ed PRs are not listed in `get_open_prs_to_fast`?

hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+    get_prs = []
+    for pr in errata.get_open_prs_to_fast(repo):
+        get_prs.append(pr)
+
+    assert get_prs == []
+
+
+def test_get_open_prs_to_fast_user_login():
+    labels_empty = []
+    prs = []
```
vrutkovs, line 439: Lets be explicit about valid and invalid input:
```
valid_prs = [
    GithubRepoPrMock(GitUser("openshift-bot"), "Enable ABC in fast channels", labels_empty
]
skip_me = [
    GithubRepoPrMock(GitUser("openshift-user"), "Enable DEF in fast channels", labels_empty,
    ...
]
all_prs = valid_prs + skip_me
...
received_prs = []
for pr in errata.get_open_prs_to_fast(repo):
    received_prs.append(pr)
# TODO: replace with more elaborate comparison which shows diff
assert received_prs == valid_prs
```

hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+
+    githubrepo = MagicMock()
+    githubtoken = MagicMock()
+    message = {
+        "errata_id" : 5678,
+    }
+
+    errata.lgtm_fast_pr_for_errata(githubrepo, githubtoken, message)
+
+    msg = "Autoapproving PR to fast after the errata has shipped\n/lgtm"
```
vrutkovs, line 597: `errata.py` needs to declare this as a var - and unittests would import that

vrutkovs (approved):
LGTM overall, lets squash all the commits and its good to go for a first attempt

hack/requirements.txt
```diff
@@ -1 +1,2 @@
 PyGithub==1.52
+pytest==6.2.2
```
wking, line 2: I personally prefer sticking to the stdlib instead of using test-only deps, because it's one less thing we might forget to bump, and the stdlib support for testing is pretty good.

hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+import os
+import re
+import tempfile
+import unittest
+import urllib
+from collections import namedtuple
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+def test_extract_errata_number_from_body_valid_url():
```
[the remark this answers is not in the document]
wking: If we drop pytest, the stdlib approach to [subtests][1] would be something like:

```python
class ExtractErrataFromBodyTest(unittest.TestCase):
    def test_valid_url(self):
        """
        Test errata number extraction from valid URLs.
        """
        for (url, expected) in [
                ('errata is https://errata.devel.redhat.com/advisory/12345', 12345):
            with self.subTest(url=url):
                self.assertEqual(errata.extract_errata_number_from_body(url), expected)
```

[1]: https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

hack/test_errata.py
wking: [`TemporaryDirectory` docs][1]:

> On completion of the context or destruction of the temporary directory object the newly created temporary directory and all its contents are removed from the filesystem.

So this should be fine as it stands without involving pytest-specific stuff.  But if we stick with pytest as the overall framework, using their tempdir fixture is probably more idiomatic than using equivalent stdlib tooling.

[1]: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

hack/test_errata.py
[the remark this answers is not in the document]
vrutkovs: That'd make test quit on first failing case, isn't it?

hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+    assert errata.extract_errata_number_from_body(body) == None
+
+
+def test_load():
+    with tempfile.TemporaryDirectory() as tempdir:
+        cachepath = os.path.join(tempdir, "cache.json")
+        cache = {"foo": "bar"}
+
+        with open(cachepath, 'w') as file:
+            json.dump(cache, file)
```
wking, line 44: I think it's probably better to test `save` and `load` as a pair, instead of inlining a `save` implementation here so you can test `load` independently.  Something like:

1. Create a temp dir
2. Try to load `cachepath`.  Confirm you get `{}`.
3. Add some dummy data.
4. `save`
5. `load`
6. Confirm you got the dummy data back unchanged.

hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+        polled_messages.append(message)
+
+    #Get arguments of urlib.request.urlopen call inside errata.poll
+    urlopen_called_with_args = urlopen_mock.call_args
+    # Get params of the url
+    url = re.search(r"call\(\'(.*)\'\)", str(urlopen_called_with_args)).group(1)
+    parsed_url = urllib.parse.urlparse(url)
+    params = urllib.parse.parse_qs(parsed_url.query)
+
+    # Assert if parametres complies with datagrepper reference
```
wking, line 94: nit `parametres` -> `parameters`.

hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+    assert int(params["page"][0]) > 0               # Page must be greater than 0
+    assert int(params["rows_per_page"][0]) <= 100   # Must be les than or equal to 100
```
wking, line 96: nit: `les` -> `less`

hack/test_errata.py
wking: That way we end up just duping the production logic in the test.  I'd rather just hard-code the expected string, instead of reformatting it from the `message`.

hack/test_errata.py
```diff
@@ -0,0 +1,598 @@
...
+    expected_messages = [x['msg'] for x in raw_messages]
+    assert expected_messages == polled_messages
+
+    # URL wasn't responsive only once, so sleep should be called only once
+    sleep_mock.assert_called_once()
+
+
+@patch("json.load")
+@patch("urllib.request.urlopen")
+def test_poll_multiple_messages(urlopen_mock, json_load_mock):
```
wking, line 207: Picking a random line so we can mark this thread resolved once it is, but one of the issues has been our regexp not matching the synopses we've seen in the wild.  Working through `git log hack/errata.py`, we want to match:

* OpenShift Container Platform 4.6 GA Images
* Moderate: OpenShift Container Platform 4.5.20 bug fix and golang security update
* Low: OpenShift Container Platform 4.3.40 security and bug fix update

while excluding:

* OpenShift Container Platform 4.5.11 optional CSI driver Operators bug fix update

Probably worth mining more of DataGrepper's history or the cache of a running errata pod for more examples to feed into the synopsis-matching test.

hack/test_errata.py
```diff
@@ -0,0 +1,738 @@
...
+        Github_mock.return_value = self.github_object_mock
+        param_list = self.prs_with_index_of_expected_pr
+
+        for (prs, message, expected_index_of_pr_to_create_issue) in param_list:
+            with self.subTest():
+                self.repo.get_pulls = MagicMock(return_value=prs)
+                errata.lgtm_fast_pr_for_errata(githubrepo, githubtoken, message)
+
+                issue_comment = prs[expected_index_of_pr_to_create_issue].create_issue_comment.call_args
+                expected_issue_comment = "Autoapproving PR to fast after the errata has shipped\n/lgtm"
```
vrutkovs, line 737:
```
if __name__ == '__main__':
    unittest.main()
```
to actually start tests

hack/test_errata.py
```diff
@@ -0,0 +1,738 @@
...
+import unittest
+import urllib
+from collections import namedtuple
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+
+GithubUserMock = namedtuple("GithubUserMock", "login")
+GithubLabelMock = namedtuple("GithubLabelMock", "name")
+class GithubPRMock:
```
vrutkovs, line 18: nitpick: two blank lines before class definition

hack/test_errata.py
```diff
@@ -0,0 +1,738 @@
...
+                and self.title == other.title   \
+                and self.labels == other.labels \
+                and self.number == other.number \
+                and self.body == other.body     \
+                and self.url == other.url       \
+                and self.html_url == other.html_url
+
+
+class ExtractErrataNumberFromBodyTest(unittest.TestCase):
+    def test_valid_url(self):
```
vrutkovs, line 42:
```suggestion
def test_url(self):
```
This test case has both valid and invalid urls, so lets have a single test for those

hack/test_errata.py
```diff
@@ -0,0 +1,738 @@
...
+            {
+                "additional_unnecessary_info": "shouldn't be processed",
+                "msg": {
+                    "errata_id": 24,
+                    "product": "SHIPPED_LIVE",
+                    "to": "RHOSE",
+                }
+            }
+        ]
+        self.multiple_raw_messages =                                    \
```
vrutkovs, line 190: A bit unreadable, maybe
```suggestion
        self.raw_messages = [
            (True, {
                    "additional_unnecessary_info": "shouldn't be processed",
                    "msg": {
                        "errata_id": 11,
                        "product": "RHOSE",
                        "to": "SHIPPED_LIVE",
                    }}),
            (True, {
                    "additional_unnecessary_info": "shouldn't be processed",
                    "msg": {
                        "errata_id": 12,
                        "product": "RHOSE",
                        "to": "SHIPPED_LIVE",
                    }}),
            (False, {
                    "additional_unnecessary_info": "shouldn't be processed",
                    "msg": {
                        "errata_id": 24,
                        "product": "SHIPPED_LIVE",
                        "to": "RHOSE",
                    }})
        ]
        self.valid_messages = [x[1] for x in self.raw_messages if x[0]]
```
would be a bit better?

vrutkovs (changes requested):

hack/test_errata.py
Davoska: I was starting the tests using `python -m unittest` in the terminal, so I missed this, thank you.

hack/test_errata.py
Davoska: I must have overlooked the backslash when I was double-checking the code, thank you.

hack/test_errata.py
Davoska: I need to be more explicit and I'll need to definitely work on naming variables, functions, etc.

By valid URL I meant URL which starts with corresponding `ERRATA_MARKER` in the `errata.py`:
`https://errata.devel.redhat.com/advisory/`

I'll change the name of the function and I'll add the tests as well.

Thanks

hack/test_errata.py
```diff
@@ -0,0 +1,738 @@
...
+        param_list = [
+            ('https://errata.devel.redhat.com/advisory/12345', 12345),
+            ('https://errata.devel.redhat.com/advisory/67890', 67890),
+            ('https://errata.devel.redhat.com/advisory/13579', 13579),
+            ('https://errata.devel.redhat.com/advisory/24680', 24680),
+            ('https://errata.devel.redhat.com/advisory/', None),
+            ('https://errata.devel.redhat.com/advisory/invalid', None)
+        ]
+        for (url, expected) in param_list:
+            with self.subTest():
```
wking, line 55: `subTest(url=url)` would [make it easier to identify the failing subtest][1], if we ever actually fail this.

[1]: https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest

hack/test_errata.py
```diff
@@ -0,0 +1,738 @@
...
+        param_list = [
+            ('http://errata.devel.redhat.com/advisory/12345', None),
+            ('https://errrata.devel.redhat.com/advisory/12345', None),
+            ('https://errata.dvel.reddhat.com/advisori/12345', None),
+            ('https://errata.devel.redhat.com/12345', None),
+            ('https://errata.devel.com/advisory/12345', None),
+            ('https://errata.redhat.com/advisory/12345', None),
+            ('https://devel.redhat.com/advisory/12345', None),
+            ('https://redhat.com/advisory/12345', None),
+            ('https://errata.com/advisory/12345', None)
```
wking, line 72: nit: no need to parameterize `expected` if it's always `None`.  Just:

```python
urls = [
    'https://...',
    ...
]
for url in urls:
    ...

File: hack/test_errata.py [hunk 19]
```diff
@@ -0,0 +1,738 @@
+import datetime
+import errata
+import github
```
wking, line 3: nit: this is not a stdlib dependency, so it should go [in it's own block after the stdlib stuff][1].

[1]: https://www.python.org/dev/peps/pep-0008/#imports

hack/test_errata.py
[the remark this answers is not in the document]
wking: Nope, a failing assert just kills that one subtest:

```console
$ git --no-pager diff -U1
diff --git a/hack/errata.py b/hack/errata.py
index 4aaaf85..7313fd0 100755
--- a/hack/errata.py
+++ b/hack/errata.py
@@ -15,3 +15,3 @@ import urllib.request
 
-import github
+#import github
 
diff --git a/hack/test_errata.py b/hack/test_errata.py
index 5a8e89a..11d197c 100644
--- a/hack/test_errata.py
+++ b/hack/test_errata.py
@@ -2,3 +2,3 @@ import datetime
 import errata
-import github
+#import github
 import http
@@ -48,5 +48,5 @@ class ExtractErrataNumberFromBodyTest(unittest.TestCase):
             ('https://errata.devel.redhat.com/advisory/67890', 67890),
-            ('https://errata.devel.redhat.com/advisory/13579', 13579),
+            ('https://errata.devel.redhat.com/advisory/13579', 'nope'),
             ('https://errata.devel.redhat.com/advisory/24680', 24680),
-            ('https://errata.devel.redhat.com/advisory/', None),
+            ('https://errata.devel.redhat.com/advisory/', 'nope again'),
             ('https://errata.devel.redhat.com/advisory/invalid', None)
$ python -m unittest
...
======================================================================
FAIL: test_valid_url (test_errata.ExtractErrataNumberFromBodyTest) (<subtest>)
Test errata number extraction from valid URLs.
----------------------------------------------------------------------
  File "/home/wking/src/openshift/cincinnati-graph-data/hack/test_errata.py", line 56, in test_valid_url
    self.assertEqual(errata.extract_errata_number_from_body(url), expected)
AssertionError: 13579 != 'nope'

======================================================================
FAIL: test_valid_url (test_errata.ExtractErrataNumberFromBodyTest) (<subtest>)
Test errata number extraction from valid URLs.
----------------------------------------------------------------------
  File "/home/wking/src/openshift/cincinnati-graph-data/hack/test_errata.py", line 56, in test_valid_url
    self.assertEqual(errata.extract_errata_number_from_body(url), expected)
AssertionError: None != 'nope again'
...
```

vrutkovs (approved):
/lgtm

vrutkovs:
/hold cancel

All comments are mostly nitpicks, we can fix it later

PR #722, opened 2021-03-24
blocked-edges/4.7: Block all 4.6 edges to 4.7.4

jottofar:


wking:
We have a 4.7.4 image now.

/retest

jottofar:
looks like we're waiting on signatures now

wking:
/retest

https://amd64.ocp.releases.ci.openshift.org/releasestream/4-stable/release/4.7.4 has been accepted, so we should have sigs.

wking:
Even if we don't have sigs yet, retest bot should bang away until it sees them:

/hold cancel

PR #726, opened 2021-03-24
build-suggestions/4.7: Bump z_min to 4.7.0

wking:
We've had a few stable releases out in 4.7.z now, so anyone still running the FCs and RCs should be able to update to one of those (e.g. 4.7.2), and then update from there on to later 4.7.z.  Raising this floor makes life easier for ART, because they attempt to validate all the update edges we bake into release images.  Raising the floor means we no longer need to test FC and RC -> 4.7.z for new releases.


LalatenduMohanty (approved):
/lgtm

PR #752, opened 2021-04-14
Cargo: bump Cincinnati

vrutkovs:
This updates graph-data.rs to latest Cincinnati, so that we could use latest tokio and reqwest crates.

/cc @PratikMahajan @wking @LalatenduMohanty


PratikMahajan (approved):
/lgtm

PR #772, opened 2021-04-23
Cargo: update dependencies

vrutkovs:
* recent Cincinnati commit
* tokio 1.5
* lazy_static 1.4.0
* reqwest 0.11

Pinned actix-http and actix-service to be compatible with Cincinnati

See #752 for previous attempt

### DIFF
--- a/graph-data.rs/Cargo.lock
+++ b/graph-data.rs/Cargo.lock
@@ -2,91 +2,71 @@
 # It is not intended for manual editing.
 [[package]]
 name = "actix-codec"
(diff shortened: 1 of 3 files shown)


wking:
Do we understand why this isn't hitting the #760 issue?  Your new pinning?

vrutkovs:
> Do we understand why this isn't hitting the #760 issue? Your new pinning?

Yup, pins prevent from pulling in unexpected actix-* crates

wking (approved):
/lgtm

PR #828, opened 2021-05-27
Bump minimum 4.7 to 4.8 requirements

sdodson:


build-suggestions/4.8.yaml
```diff
@@ -2,10 +2,10 @@
 # If you specify an arch it must have the full set of values
 ---
 default:
-  minor_min: 4.7.0
+  minor_min: 4.7.11
```
LalatenduMohanty, line 5: I think we should use 4.7.7 as that's the version we had the stable edge from 4.6.

LalatenduMohanty (changes requested):

build-suggestions/4.8.yaml
sdodson: updated to set it at 4.7.7

LalatenduMohanty (approved):
/lgtm

PR #838, opened 2021-06-04
Block 4.7.11 due to bug 1953518

vrutkovs:
CMO goes degraded when user monitoring workflow is enabled due to DNS changes https://bugzilla.redhat.com/show_bug.cgi?id=1953518

### DIFF
--- a/blocked-edges/4.7.11.yaml
+++ b/blocked-edges/4.7.11.yaml
@@ -1,0 +1,3 @@
+to: 4.7.11
+from: .*
+# CMO goes degraded when user monitoring workflow is enabled due to DNS changes https://bugzilla.redhat.com/show_bug.cgi?id=1953518


blocked-edges/4.7.11.yaml
```diff
@@ -0,0 +1,3 @@
+to: 4.7.11
+from: .*
+# CMO goes degraded when user monitoring workflow is enabled due to DNS changes https://bugzilla.redhat.com/show_bug.cgi?id=1953518
```
LalatenduMohanty, line 3: 4.7.11 has the fix, so this is not correct. We need to block edges to versions before 4.7.11

LalatenduMohanty (changes requested):

blocked-edges/4.7.11.yaml
LalatenduMohanty: Not all versions before 4.7.11, so trying to find out which version.

sdodson:
For anyone interested we're searching for instances of this bug affecting 4.7 clusters because while we the same change that fixed the problem in 4.6.31 went into 4.7.11 we have no observed cases of the problem occuring in 4.7.z. Nor do we have root cause for what seemingly introduced the problem in 4.6.30.

sdodson:
/close
The issue present in 4.6.30 is not present in any 4.7 release because the DNS service change and the CMO change to accommodate that both went into the same release, 4.7.11.

PR #840, opened 2021-06-05
hack/errata: Match the 4.7.13 errata message

wking:
It has a slightly different bug/security phrasing than previous bug/security errata.  Also add a helper to treat RHBA and RHSA as synonyms, because ART sometimes guesses wrong when baking the URI into the release image (details in the commit messages).


LalatenduMohanty (approved):
/lgtm

PR #871, opened 2021-06-17
Tombstone 4.7.17

vrutkovs:
MCO does not cordon nodes on upgrade https://bugzilla.redhat.com/show_bug.cgi?id=1973006


wking (approved):
/lgtm

PR #875, opened 2021-06-22
Align minimum 4.7.z to latest at code freeze

sdodson:
z_min will of course be updated at GA but minor_min will not be updated again unless we become aware of a specific identified issue affecting upgrade success


sdodson:
/hold
Lets wait until the next rc is built before we merge this.

PR #897, opened 2021-06-29
Update minimum 4.7.z to 4.8

sdodson:
On account of https://bugzilla.redhat.com/show_bug.cgi?id=1977383
which leads to potential workload disruption during upgrades to
4.8. Note, this is only the assumed target version, this may change.


LalatenduMohanty (approved):
/lgtm

PR #926, opened 2021-07-17
hack/stabilization-changes: Automatically follow upstream before new poll rounds

wking:
In single-shot mode, assume the caller has the local directory where they want it to be.  For poll-mode, do that for the first round, although there's a chance we've gone a bit stale since the image was built (in which case we might fail while trying to recreate an existing upstream branch, but no big deal).  But after waking up from the poll-sleep for the next attempt, fetch and checkout the upstream tip, to ensure we stay up to date.  At least with the graph-data; we won't be picking up new Python code this way.

I'd somewhat prefer to [trigger our BuildConfig with a GitHub webhook][1] to get this same effect (rebuilding based on the upstream tip, and then bumping the deployment to run that new build).  That would get new Python picked up too.  But we can't do that conveniently today, because we're deploying the cluster inside Red Hat's VPN, where GitHub can't reach the cluster to poke the BuildConfig trigger.  One benefit to this commits internal tracking is that we preserve the `next_notification` time, so we don't flood the webhook channel with waiting notifications.

[1]: https://docs.openshift.com/container-platform/4.7/cicd/builds/triggering-builds-build-hooks.html#builds-using-image-change-triggers_triggering-builds-build-hooks

### DIFF
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -159,10 +159,10 @@ def load_channels(directory):
 
 def get_promotions(path):
     # https://git-scm.com/docs/git-blame#_the_porcelain_format
-    process = subprocess.run(['git', 'blame', '--first-parent', '--porcelain', path], check=True, capture_output=True)
+    process = subprocess.run(['git', 'blame', '--first-parent', '--porcelain', path], check=True, capture_output=True, text=True)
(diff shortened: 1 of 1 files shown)


vrutkovs (approved):
/lgtm

PR #938, opened 2021-07-22
channels/candidate-4.8: Tombstone 4.8.1

wking:
s390x missed a kernel bump that went into the other platforms.  We're building 4.8.2 to catch s390x up.


LalatenduMohanty (approved):
/lgtm

wking:
registry.redhat.io is having DNS issues at the moment, and I'm confident in this change.

/override ci/prow/e2e
/override ci/prow/images

PR #943, opened 2021-07-23
hack/stabilization-changes: Use the configured token to push

wking:
Avoid:

```
Branch promote-4.6.39-to-stable-4.6 set up to track remote branch master from origin.
[promote-4.6.39-to-stable-4.6 71a3666] channels/stable-4.6: Promote 4.6.39
 1 file changed, 1 insertion(+)
fatal: could not read Username for 'https://github.com': No such device or address
ERROR:   failed to promote 4.6.39 to stable-4.6: Command '['git', 'push', '-u', 'origin', 'promote-4.6.39-to-stable-4.6']' returned non-zero exit status 128.
```

I think that's Git saying "I need to authenticate this push.  Let me ask for credentials using the TTY.  What!?  No TTY in this container!  Aaaah!".

I spent a while looking at [`git-credentials(7)`][1], which sounded promising, but while there is clear discussion on how to set up the cache helper and such, and how to insert creds into the cache via TTY prompts, I couldn't find a way to insert credentials without using a TTY.

So instead, this commit constructs a remote upstream URI, inserting the configured GitHub token as [user information][2].  And then we push to that URI instead of pushing to the remote.  The token will be visible to folks with `/proc/` access, but they could already read it out of our environment variables.  And we keep it off the local disk.

[`urlsplit` returns a named tuple][3], and [the `_replace` method][4] allows us to slot in our token user information.

[1]: https://git-scm.com/docs/gitcredentials
[2]: https://datatracker.ietf.org/doc/html/rfc3986/#section-3.2.1
[3]: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit
[4]: https://docs.python.org/3/library/collections.html#collections.somenamedtuple._replace

### DIFF
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -317,6 +317,16 @@ def notify(message, webhook=None):
 def promote(version, channel_name, channel_path, subject, body, github_repo, github_token, upstream_remote, upstream_branch):
     if not github_token:
         raise ValueError('cannot promote without a configured GitHub token')
+
+    process = subprocess.run(['git', 'remote', 'get-url', '--push', upstream_remote], check=True, capture_output=True, text=True)
+    upstream_uri = urllib.parse.urlsplit(process.stdout.strip())
+    if upstream_uri.scheme != 'https':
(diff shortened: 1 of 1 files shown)


vrutkovs (approved):
/lgtm

PR #960, opened 2021-08-03
Extend promotion delays

sdodson:
The most current minor goes to 96hr, and in order to avoid the previous
minor not having upgrade paths into the latest minor extend its delay to
reduce the likelihood of that happening. This may be moot until we actually
bring stable channel upgrades to 4.8 and perhaps by the time we achieve
that milestone we decide to get closer to 72, 48, 24 across 4.N, 4.N-1, 4.N-2.

/cc @wking @LalatenduMohanty

### DIFF
--- a/channels/stable-4.7.yaml
+++ b/channels/stable-4.7.yaml
@@ -1,5 +1,5 @@
 feeder:
-  delay: PT48H
+  delay: PT72H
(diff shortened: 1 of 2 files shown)


LalatenduMohanty (approved):
/lgtm

PR #987, opened 2021-08-16
Block edges from 4.7.z to 4.8.{2-4}

cblecker:
Blocking 4.7.z to 4.8.{2-4} edges due to BZ1988576.

### DIFF
--- a/blocked-edges/4.8.2.yaml
+++ b/blocked-edges/4.8.2.yaml
@@ -1,0 +1,3 @@
+to: 4.8.2
+from: 4\.7\..*
+# Authentication operator fails to become available during upgrade https://bugzilla.redhat.com/show_bug.cgi?id=1988576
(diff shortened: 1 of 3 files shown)


sdodson:
/hold
Awaiting confirmation that upgrading to 4.8.5 heals affected clusters or not.

blocked-edges/4.8.0.yaml
```diff
@@ -0,0 +1,3 @@
+to: 4.8.0
+from: 4\.7\..*
+# Authentication operator fails to become available during upgrade https://bugzilla.redhat.com/show_bug.cgi?id=1988576
```
wking, line 3: 4.8.0 was never supported (#918):

```console
$ curl -s 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=fast-4.8' | jq -r '.nodes[].version' | sort -V | tail -n5
4.7.23
4.8.2
4.8.3
4.8.4
4.8.5
```

I don't think we block updates for unsupported targets?

/hold

wking: Fixed in 2297eae.

/hold cancel
/lgtm

PR #1008, opened 2021-08-25
Block edges from 4.7.z to 4.7.24 and 4.8.z

PratikMahajan:
Blocking edges due to BZ1987108


blocked-edges/4.8.5.yaml
```diff
@@ -0,0 +1,3 @@
+to: 4.8.5
+from: 4\.7\..*
+# SDN is down and kube-apiserver cannot reach aggregated apiserver: no route to host, SDN Packet loss resulting in service unavailability.  https://bugzilla.redhat.com/show_bug.cgi?id=1987108
```
LalatenduMohanty, line 3: All 4.8.2+ are known to be affected because of the bugs, so we also need to add this bug to previous blocked edges

LalatenduMohanty (changes requested):

blocked-edges/4.8.5.yaml
PratikMahajan: Added

blocked-edges/4.7.24.yaml
```diff
@@ -0,0 +1,3 @@
+to: 4.7.24
+from: 4\.7\..*
+# SDN is down and kube-apiserver cannot reach aggregated apiserver: no route to host, SDN Packet loss resulting in service unavailability.  https://bugzilla.redhat.com/show_bug.cgi?id=1987108
```
LalatenduMohanty, line 3: We should change the bug description to `Networking issue with vSphere clusters running HW14 and later`

PratikMahajan: Changes the bug description!

blocked-edges/4.7.24.yaml
```diff
@@ -0,0 +1,3 @@
+to: 4.7.24
+from: 4\.7\..*
```
wking, line 2: /hold

I don't understand why 4.6 -> 4.7.24 is ok.

sdodson: Yeah we need to block that too, sorry I forgot about that when proposing this. So everything to 4.7.24

PR #1012, opened 2021-08-27
channels/candidate-4.[78]: Tombstone 4.7.26 and 4.8.7

wking:
These are both vulnerable to [the vSphere networking bug][1].  They could still be useful updates from other versions which are also vulnerable (because vulnerable -> vulnerable is not a regression), but we're pushing 4.7.27 and 4.8.8 out soon with fixes, so tombstoning because we don't need to clutter the supported graph.

I'm also dropping the 4.8.7 edge blocks, because we don't manage edges for candidate-only releases.  If folks want to send candidate-level clusters across those edges, despite them being unsupported, they might discover further issues, which we can use to protect supported clusters taking other, similar edges.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1987108


wking:
I'd forgotten about `candidate-4.9` here, and Scott is recovering in #1134 now that the `fast-4.9` `filter` will no longer be holding 4.8.7 out.

PR #330, opened 2020-07-14
OWNERS: OTA may now approve all graph changes

sdodson:

PR #1040, opened 2021-09-03
OWNERS: Drop release-admins

wking:
I'm pretty sure this is what @sdodson meant to do in 7dd81a947d (#330), and he just didn't think to look in the `raw/` directory.

This also saves us from having to attempt to maintain a release-admins alias (e.g. Alex no longer works for Red Hat, and none of the others have weighed in on graph-data changes in a while now).  Most of the folks who used to be in release-admins have direct access to manipulate GitHub labels anyway, so they don't need to go through Prow to drop approval labels on anything; they can just do that directly via GitHub actions.


LalatenduMohanty (approved):
/lgtm

PR #1048, opened 2021-09-08
channels/stable-4.8: Promote 4.8.10

wking:
The bot would prefer to wait longer after f8104a49d4 (#1044), but because the CRI-O /run leak that 4.8.10 fixes has such serious potential impact on all clusters, and the existing 4.8.10 Telemetry/Insights feedback looks positive, we want to push it into stable-4.8 without waiting for the full, configured `delay` to pass.


LalatenduMohanty (approved):
/lgtm

PR #1075, opened 2021-09-20
channels/stable-4.8: Remove filter to include 4.7.z

wking:
Generated with:

```console
$ sed -i '/^  filter: /d' channels/stable-4.8.yaml
```

but using the still-in-flight version of that script from #1039 to include the promotions in this commit, so the stabilization bot doesn't follow up with a whole bunch of parallel promotion pull requests.

We've been cooking in fast-4.8 long enough that we feel fairly confident that major 4.7 -> 4.8 issues will have surfaced by now.


wking:
Waiting for @sdodson to sign off.

/hold

PR #1076, opened 2021-09-20
hack/validate-blocked-edges: Add basic blocked-edges validation

wking:
To keep us from forgetting to set `url`, etc.

We'll probably [want to have PromQL parser validation later][1], but I write Python a lot faster than I write Go.  This should avoid obvious mistakes while the PromQL parser gets wired up to a validator.

[1]: https://github.com/openshift/enhancements/blob/4412b51182ba6e007f1308a73e290e81fd66d092/enhancements/update/targeted-update-edge-blocking.md#promql-validation

### DIFF
--- a/hack/validate-blocked-edges.py
+++ b/hack/validate-blocked-edges.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+
+import os
+
+import yaml
+
+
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (approved):
/lgtm

PR #1078, opened 2021-09-21
Revert "blocked-edges/4.7.4*: Targeted edge blocking and version 1.1.0"

sdodson:
Reverts openshift/cincinnati-graph-data#1056

This appears to have broken graph building.
```
2021-09-21T13:09:26Z ERROR graph_builder::graph] unrecognized graph-data version 1.1.0
    ; supported versions: ["1.0.0"]
```

PR #1056, opened 2021-09-10
blocked-edges/4.7.4*: Targeted edge blocking and version 1.1.0

wking:
So we can explain why we're blocking the different edges: openshift/enhancements#821.

The `zz` in the filename for the vSphere hostname block ensures that one sorts last alphabetically, because it's the broadest block, and legacy Cincinnati services will prefer the final regular expression they load for a given `to` target.

PR #1089, opened 2021-09-23
blocked-edges/4.7.4*: Targeted edge blocking and version 1.1.0

wking:
So we can explain why we're blocking the different edges openshift/enhancements#821 (the `promql` -> `PromQL` type change is in flight with openshift/enhancements#910).

The `zz` in the filename for the vSphere hostname block ensures that one sorts last alphabetically, because it's the broadest block, and legacy Cincinnati services will prefer the final regular expression they load for a given `to` target.

This is basically a second attempt at my earlier 39bc2fb686 (#1056), which ended up getting reverted in da1254a39f (#1078) because the production Cincinnati was mad about the 1.1.0 version string.  openshift/cincinnati#538 taught Cincinnati to relax, and now that's live (and we never shipped any versions that would be mad about 1.1.0 to customers, the 4.6.0 Update Service operator went out before openshift/cincinnati#314).


LalatenduMohanty (approved):
/lgtm

PR #1117, opened 2021-10-08
channels/stable-4.*: Fix delay "PT7D" to "P1W"

wking:
"one week" puts all the weight on the unit, while "seven days" requires you to think about both the unit and the magnitude ;).

But more importantly, `T7D` is illegal, because [`T` is the marker for "and now we're going to start talking about times"][1], and days are larger than times.

[1]: https://datatracker.ietf.org/doc/html/rfc3339#page-13

### DIFF
--- a/channels/stable-4.6.yaml
+++ b/channels/stable-4.6.yaml
@@ -1,5 +1,5 @@
 feeder:
-  delay: PT7D
+  delay: P1W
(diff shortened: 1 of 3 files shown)


wking:
I'm confident enough in this to self-lgtm, since the other approvers are all clocked out by now.

PR #1132, opened 2021-10-18
Dockerfile: install Rust in UBI

vrutkovs:
Use UBI container when building graph-data.rs so that fresh Rust version could be used


LalatenduMohanty (approved):
/lgtm

PR #1134, opened 2021-10-18
4.9 GA release activities

sdodson:
Fill in 4.8.z into fast-4.9
Add 4.9.0 to stable-4.9
Update promotion automation

PR #1180, opened 2021-11-02
hack/stabilization-changes.py: link to commit instead of hash/message, update formatting

vrutkovs:
Show a link to commit instead of hash+commit message, convert dates to human readable (without seconds or milliseconds) and show expected time to merge for promotions.

This changes the webhook bot text from
`* Recommend waiting to promote 4.6.49 to fast-4.6; it was promoted the feeder candidate-4.6 by 7f82da1c9c (Merge pull request #1169 from openshift/pr-candidate-4.6.49, 2021-10-27, 5 days, 20:28:09.641705) https://access.redhat.com/errata/RHBA-2021:4009 is not public.`
to
`* Recommend waiting to promote 4.6.49 to fast-4.6; it was promoted the feeder candidate-4.6 by <https://github.com/openshift/cincinnati-graph-data/commits/7f82da1c9c|7f82da1c9c> Merged 5 days, 20 hours, 27 minutes ago) https://access.redhat.com/errata/RHBA-2021:4009 is not public.`
for non-public erratas 

and  
`* Recommend waiting to promote 4.7.36 to stable-4.7; it was promoted the feeder fast-4.7 by 1843c9c1ac (Merge pull request #1164 from openshift/promote-4.7.36-to-fast-4.7, 2021-10-27, 6 days, 0:58:18.461610)`
to
`* Recommend waiting to promote 4.7.36 to stable-4.7; it was promoted the feeder fast-4.7 by <https://github.com/openshift/cincinnati-graph-data/commits/1843c9c1ac|1843c9c1ac> Expected to be promoted on 2021-11-03 10:55:55)`
for delayed promotions.

Format `<https://foo.com|bar>` should be translated by Slack to a hyperlink

### DIFF
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -151,7 +151,7 @@ def stabilize_release(version, channel_name, channel_path, delay, errata, feeder
                 time_message = "Current promotion delay will be satisfied on {}".format(feeder_promotion['committer-time'] + delay)
             else:
                 time_message = "Merged {}".format(merged_ago)
-            yield 'Recommend waiting to promote {} to {}; it was promoted the to feeder {} by {} {}){}'.format(
+            yield 'Recommend waiting to promote {} to {}; it was promoted the to feeder {} by {} {}) {}'.format(
                 version,
                 channel_name,
                 feeder_name,
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -105,11 +105,11 @@ def stabilize_release(version, channel_name, channel_path, delay, errata, feeder
             public_errata_message = ' {} is{} public.'.format(errata_uri, '' if errata_public else ' not')
     if (delay is not None and version_delay > delay) or errata_public:
         path_without_extension, _ = os.path.splitext(channel_path)
+        commit_link = '<https://github.com/openshift/cincinnati-graph-data/commits/{0}|{0}>'.format(feeder_promotion['hash'][:10])
         subject = '{}: Promote {}'.format(path_without_extension, version)
-        body = 'It was promoted to the feeder {} by {} ({}, {}) {} ago.{}'.format(
+        body = 'It was promoted to the feeder {} by {} ({}) {} ago.{}'.format(
                 feeder_name,
-                feeder_promotion['hash'][:10],
-                feeder_promotion['summary'],
+                commit_link,
                 feeder_promotion['committer-time'].date().isoformat(),
                 version_delay,
                 public_errata_message,
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -147,10 +147,10 @@ def stabilize_release(version, channel_name, channel_path, delay, errata, feeder
             time_message = None
             if delay:
                 # Show expected date to promote
-                time_message = "Expected to be promoted on {}".format(feeder_promotion['committer-time'] + delay)
+                time_message = "Current promotion delay will be satisfied on {}".format(feeder_promotion['committer-time'] + delay)
             else:
                 time_message = "Merged {}".format(merged_ago)
-            yield 'Recommend waiting to promote {} to {}; it was promoted the feeder {} by {} {}){}'.format(
+            yield 'Recommend waiting to promote {} to {}; it was promoted the to feeder {} by {} {}){}'.format(
                 version,
                 channel_name,
                 feeder_name,
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -2,6 +2,7 @@
 
 import codecs
 import datetime
+import dateutil.relativedelta
 import http
 import json
 import logging
@@ -94,24 +95,33 @@ def stabilize_channel(name, channel, channels, channel_paths, **kwargs):
             **kwargs)
 
 
+def human_readable_relativedelta(delta):
+    attrs = ['years', 'months', 'days', 'hours', 'minutes']
+    human_readable = ['%d %s' % (getattr(delta, attr), attr if getattr(delta, attr) > 1 else attr[:-1])
+        for attr in attrs if getattr(delta, attr)]
+    return "{} ago".format(", ".join(human_readable))
+
+
 def stabilize_release(version, channel_name, channel_path, delay, errata, feeder_name, feeder_promotion, cache, waiting_notifications=True, **kwargs):
     now = datetime.datetime.now()
     version_delay = now - feeder_promotion['committer-time']
+    merged_ago = human_readable_relativedelta(
+        dateutil.relativedelta.relativedelta(now, feeder_promotion['committer-time']))
     errata_public = False
     public_errata_message = ''
+    commit_link = '<https://github.com/openshift/cincinnati-graph-data/commits/{0}|{0}>'.format(feeder_promotion['hash'][:10])
     if errata:
         errata_uri, errata_public = public_errata_uri(version=version, channel=feeder_name, cache=cache)
         if errata_uri:
             public_errata_message = ' {} is{} public.'.format(errata_uri, '' if errata_public else ' not')
     if (delay is not None and version_delay > delay) or errata_public:
         path_without_extension, _ = os.path.splitext(channel_path)
-        commit_link = '<https://github.com/openshift/cincinnati-graph-data/commits/{0}|{0}>'.format(feeder_promotion['hash'][:10])
         subject = '{}: Promote {}'.format(path_without_extension, version)
-        body = 'It was promoted to the feeder {} by {} ({}) {} ago.{}'.format(
+        body = 'It was promoted to the feeder {} by {} ({}) {}.{}'.format(
                 feeder_name,
                 commit_link,
                 feeder_promotion['committer-time'].date().isoformat(),
-                version_delay,
+                merged_ago,
                 public_errata_message,
             )
         try:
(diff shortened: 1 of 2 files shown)


hack/stabilization-changes.py
```diff
@@ -2,6 +2,7 @@
 
 import codecs
 import datetime
+import dateutil.relativedelta
```
wking, line 5: [`dateutil`][1] is not part of Python's standard library.  I'm not going to attempt local implementations of YAML parsers or GitHub clients, and we can't avoid those for this script, so I'm fine with 3rd party libraries for those.  But I think we can get by without convenient duration formating, when Python's stdlib formatting isn't too bad:

```console
$ python -c 'import datetime; print(datetime.timedelta(seconds=876543.21))'
10 days, 3:29:03.210000
```

If we want to round to the nearest second to avoid milliseconds, I'd be fine with:

```python
version_delay = datetime.timedelta(seconds=round(now - feeder_promotion['committer-time']).total_seconds()))
```

like:

```console
$ python -c 'import datetime; print(datetime.timedelta(seconds=round(datetime.timedelta(seconds=876543.21).total_seconds())))'
10 days, 3:29:03
```

Thoughts?

[1]: https://pypi.org/project/python-dateutil/

hack/stabilization-changes.py
```diff
@@ -94,24 +95,33 @@ def stabilize_channel(name, channel, channels, channel_paths, **kwargs):
...
         if errata_uri:
             public_errata_message = ' {} is{} public.'.format(errata_uri, '' if errata_public else ' not')
     if (delay is not None and version_delay > delay) or errata_public:
         path_without_extension, _ = os.path.splitext(channel_path)
         subject = '{}: Promote {}'.format(path_without_extension, version)
-        body = 'It was promoted to the feeder {} by {} ({}, {}) {} ago.{}'.format(
+        body = 'It was promoted to the feeder {} by {} ({}) {}.{}'.format(
                 feeder_name,
-                feeder_promotion['hash'][:10],
-                feeder_promotion['summary'],
```
wking, line 112: This body is used for the commit message too:

```console
$ git --no-pager log --no-merges -1 channels/fast-4.*
commit 527f57a94f5fba118a5f256e9bbf2fb0e00f6fe2 (origin/promote-4.8.18-to-fast-4.9, origin/pr/1182)
Author: Automated Stabilization Robot <noreply@redhat.com>
Date:   Tue Nov 2 14:51:24 2021 +0000

    channels/fast-4.9: Promote 4.8.18
    
    It was promoted to the feeder candidate-4.9 by 885ef69e5e (Merge pull
    request #1177 from openshift/pr-candidate-4.8.18, 2021-10-28) 5 days,
    3:08:01.003849 ago. https://access.redhat.com/errata/RHBA-2021:4020 is
    public.
```

And while I don't mind HTTPS commit message links in the Slack notifications, I'd rather stick to the `%h (%s, %d)` format like:

```console
$ git --no-pager log --format=reference -1 885ef69e5e
885ef69 (Merge pull request #1177 from openshift/pr-candidate-4.8.18, 2021-10-28)
```

See [the kernel docs][1] (which do not include the date suggestion), [the Git docs][2], and git/git@1f0fc1db8599f87520494ca4f0e3c1b6fabdf997.

I'm open to formatting different bodies for Slack and the commit messages, but I'm also fine sticking with the current bias towards the read-forever commit messages, even if that makes the Slack messages a bit chatty.

[1]: https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes
[2]: https://github.com/git/git/blob/0cddd84c9f3e9c3d793ec93034ef679335f35e49/Documentation/SubmittingPatches#L146-L167

hack/stabilization-changes.py
```diff
@@ -128,16 +138,21 @@ def stabilize_release(version, channel_name, channel_path, delay, errata, feeder
...
     else:
-        _LOGGER.info('  waiting: {} ({}){}'.format(version, version_delay, public_errata_message))
+        _LOGGER.info('  waiting: {} ({}){}'.format(version, merged_ago, public_errata_message))
         if waiting_notifications:
-            yield 'Recommend waiting to promote {} to {}; it was promoted the feeder {} by {} ({}, {}, {}){}'.format(
+            # Show either expected promotion time or when
+            time_message = None
+            if delay:
+                # Show expected date to promote
+                time_message = "Expected to be promoted on {}".format(feeder_promotion['committer-time'] + delay)
```
wking, line 147: I'm fine with this, but "expected" might be too strong.  Perhaps, `Current promotion delay will be satisfied on {}` or something that is more clearly "I'm just talking about my current config" and has less likelihood of being treated as a prediction of future activity (because the config could change, or we could [promote by hand without waiting out the delay][1], etc.).

[1]: https://github.com/openshift/cincinnati-graph-data/pull/1134/files#diff-4fda33ea6ee80e385cd006122ed0861e642f546da009afc6a3c8d38ef226d233R7

hack/stabilization-changes.py
```diff
@@ -128,16 +138,21 @@ def stabilize_release(version, channel_name, channel_path, delay, errata, feeder
         else:
             yield '{}. {} {}'.format(subject, body, pull.html_url)
     else:
-        _LOGGER.info('  waiting: {} ({}){}'.format(version, version_delay, public_errata_message))
+        _LOGGER.info('  waiting: {} ({}){}'.format(version, merged_ago, public_errata_message))
         if waiting_notifications:
-            yield 'Recommend waiting to promote {} to {}; it was promoted the feeder {} by {} ({}, {}, {}){}'.format(
```
wking, line 133: I'd missed a `to` (should be `...promoted to the feeder...`).  If that text survives this PR, we should also fix the missing `to`.

hack/stabilization-changes.py
vrutkovs: Do we need commit messages though? These are autogenerated and don't carry any additional useful info imo.

hack/stabilization-changes.py
vrutkovs: Good idea, reworked in 063f073faa0bacb073757f6d35e135bc40b19cfd

hack/stabilization-changes.py
vrutkovs: Right, actually we don't need `dateutil`. Reworked this in aa7cd1d6dc4c984a3df6b9b082b47361af97d093

hack/stabilization-changes.py
```diff
@@ -94,24 +94,37 @@ def stabilize_channel(name, channel, channels, channel_paths, **kwargs):
             **kwargs)
 
 
+def human_readable_relativedelta(delay):
+    delta = {}
```
wking, line 98: your later `keys()` iteration seems like you want to use [`OrderedDict()`][1] here.

[1]: https://docs.python.org/3/library/collections.html#collections.OrderedDict

hack/stabilization-changes.py
wking: The referenced PR in the merge-commit's subject seems useful to me.  And having the inlined message from the referenced commit just makes reading it friendlier than an abbreviated commit hash and date, which... might be a useful sanity check?  Also, a Slack-formatted link dropping in `<https://github.com/openshift/cincinnati-graph-data/commits/{0}|{0}>` seems like distracting noise in the commit message.

hack/stabilization-changes.py
```diff
@@ -128,16 +141,21 @@ def stabilize_release(version, channel_name, channel_path, delay, errata, feeder
...
         if waiting_notifications:
-            yield 'Recommend waiting to promote {} to {}; it was promoted the feeder {} by {} ({}, {}, {}){}'.format(
+            # Show either expected promotion time or when
+            time_message = None
+            if delay:
+                # Show expected date to promote
+                time_message = "Current promotion delay will be satisfied on {}".format(feeder_promotion['committer-time'] + delay)
+            else:
+                time_message = "Merged {}".format(merged_ago)
+            yield 'Recommend waiting to promote {} to {}; it was promoted the to feeder {} by {} {}){}'.format(
```
wking, line 153: I think you need `)  {}` at the end or some other way to get some whitespace in front of the text in `time_message`

wking: also `the to feeder` -> `to the feeder`

hack/stabilization-changes.py
```diff
@@ -128,16 +142,21 @@ def stabilize_release(version, channel_name, channel_path, delay, errata, feeder
...
         if waiting_notifications:
-            yield 'Recommend waiting to promote {} to {}; it was promoted the feeder {} by {} ({}, {}, {}){}'.format(
+            # Show either expected promotion time or when
+            time_message = None
+            if delay:
+                # Show expected date to promote
+                time_message = "Current promotion delay will be satisfied on {}".format(feeder_promotion['committer-time'] + delay)
+            else:
+                time_message = "Merged {}".format(merged_ago)
+            yield 'Recommend waiting to promote {} to {}; it was promoted the to feeder {} by {} {}) {}'.format(
```
wking, line 154: This is still missing a period before the start of the `time_message` sentence, and it includes an unbalanced `)` left over from wrapping the old summary/date.  I still like the `feeder {} by {}. {}{}` from [here][1].

[1]: https://github.com/openshift/cincinnati-graph-data/pull/1180#discussion_r742212147

hack/stabilization-changes.py
vrutkovs: Fixed in 34a8efcd0e9cb5c179a6d492c71fe33ee4a495bc

hack/stabilization-changes.py
vrutkovs: Sure, added summary and merge date back

PR #1203, opened 2021-11-10
channels/eus-4.8: Promote 4.6.49

wking:
It was promoted to stable-4.7 by 38910847ce (#1202).  As discussed in 0e42d67f0b (#1194), we don't have a clean way to auto-promote 4.6.z into eus-4.8 at the moment, so I'm promoting it by hand.


LalatenduMohanty (approved):
/lgtm

PR #1275, opened 2021-12-09
hack/show-edges: Teach to walk internal-channels

wking:
Because showing graphs for those can be convenient too.  Catches up with stabilization-changes pivots in bda213e3ef (#112).


LalatenduMohanty (approved):
/lgtm

PR #1281, opened 2021-12-11
channels/candidate-4.10: Create this channel

wking:
We're approaching feature freeze for 4.10, and ART will soon begin cutting 4.10 feature candidates and adding them to `candidate`.  Create a 4.10-pinned channel to share these externally, so folks can start testing 4.9 -> 4.10-fc and 4.10-fc -> 4.10.fc' edges.

I'm not creating `fast-4.10` or later yet, because we're still far from GA, and we don't want folks running 4.9.z to see a fast-4.10 channel, get all excited, switch to that channel, and then not see any 4.10.z targets for weeks.


LalatenduMohanty (approved):
/lgtm

PR #1346, opened 2022-01-11
OTA-509: Add a test to verify that all stable releases have errata link

vrutkovs:
This commit adds a new test to ensure that all released versions have valid errata links:
* Fetch all parsed releases in all channels
* Extract predicate from version. If its not empty it should match arch suffix (arm64/amd64), if its not this release is omitted from the check
* If the version without a predicate matches a known faulty version its being omitted

The test verifies the following things:
* Release can be converted into `ConcreteRelease`
* Release has "url" metadata key
* URL in the key matches errata URL regex


LalatenduMohanty (approved):
/lgtm

PR #1394, opened 2022-01-19
hack/generate-weekly-report: Automate some busywork

wking:
Run like:

```console
$ hack/generate-weekly-report.py --stats https://ota-reports.example.com/.../stats.html ac07fe96e8bf19551102c2a651731415e836ac72
```

Paste into the code-edit view of the internal blog, edit/extend as needed, and publish!

Also factors some shared code out into `util.py`, as described in 59d8ff6da0fe53.

### DIFF
--- a/hack/generate-weekly-report.py
+++ b/hack/generate-weekly-report.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python3
+
+import codecs
+import datetime
+import re
+import ssl
+import subprocess
+import urllib.request
+
+import util
+
+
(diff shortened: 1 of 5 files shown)


wking:
I pushed c8cceb7 -> 4261288, rebasing onto `master` and adding `--final-commit`, so folks who want to compare with `origin/master` or whatever can do that without having it checked out:

```console
$ ./hack/generate-weekly-report.py --help | grep final-commit
usage: generate-weekly-report.py [-h] [--final-commit REF] [--stats URI]
  --final-commit REF  Git reference for the final commit. (default: HEAD)
```

PR #1401, opened 2022-01-20
A prerequsite for 4.9 to 4.10 upgrades has been identified

sdodson:
https://bugzilla.redhat.com/show_bug.cgi?id=2037665

While 4.9.18 includes logic to ensure that clusters cannot upgrade
to 4.10 without addressing the pre-requisite it introduces a minor
regression where it incorrectly blips through Upgradeable=False. As
such we will file a subsequent bump to a build with the complete
set of changes when that 4.9.z is tagged. Bumping now will avoid
potential known issues in upgrading between 4.9 and 4.10, so we'll
bump twice.

### DIFF
--- a/build-suggestions/4.10.yaml
+++ b/build-suggestions/4.10.yaml
@@ -2,7 +2,7 @@
 # If you specify an arch it must have the full set of values
 ---
 default:
-  minor_min: 4.9.0
+  minor_min: 4.9.18
   minor_max: 4.9.9999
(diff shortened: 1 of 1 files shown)


sdodson:
/hold
Wait until 4.9.18 is tagged and available in candidate channel
/cc @ingvagabund

wking (approved):
/lgtm

PR #1437, opened 2022-01-31
blocked-edges/4.10.0-fc.2-modified-aws-load-balancer-service: Example conditional update

wking:
Since 39bc2fb686 (#1056), we've had some example conditional updates back on `* -> 4.7.4`.  But now that 4.10 has feature candidates, this commit adds a conditional edge based on a minor 4.10.0-fc.2 issue that was fixed in fc.3.  Folks interested in seeing conditional edges in action can install fc.0 or fc.1 on AWS, and see that this conditional edge is not recommended.  Or they can install fc.0 or fc.1 on a non-AWS platform to see that the conditional edge is recommended.


vrutkovs (approved):
/lgtm

PR #1518, opened 2022-02-17
Block 4.10 upgrades from less than 4.9.19

sdodson:
While we raised the minimums someone running versions less than the new minimum may see available 4.10 upgrades in the candidate channel and take them. We'd prefer everyone meet the current minimum version requirements so that we're not chasing down any bugs from those versions.

I can into this because I installed 4.9.18 and it offered me upgrades to 4.10.0-rc.0.


wking (approved):
This diverges from our policy of not blocking candidate-only edges (e.g. #127), but I'm fine landing this while we sort out a more nuanced policy.

/lgtm

PR #1610, opened 2022-03-10
internal-channels/fast: Promote 4.10.3

vrutkovs:
It was promoted to the feeder candidate by 21db2f2011 (Merge pull request #1592 from openshift/pr-candidate-4.10.3, 2022-03-08) 2 days, 7:57:36.589396 ago. https://access.redhat.com/errata/RHSA-2022:0056 is public.

PR #1594, opened 2022-03-08
Adding fast and stable channel for 4.10 release

LalatenduMohanty:
Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

PR #1611, opened 2022-03-10
channels: Create fast-4.10 and stable-4.10

wking:
Now that we're GA, a068674144 (#1610).


wking:
Landing this same diff via #1594.

/close

PR #1613, opened 2022-03-10
hack/release-ga: Automate GA steps

wking:
Make it harder to forget something.  I've left of the final:

```console
$ git add channels/*
$ git commit
```

but we could do that in the future if we wanted to control the commit message.

### DIFF
--- a/hack/release-ga.sh
+++ b/hack/release-ga.sh
@@ -1,0 +1,76 @@
+#!/bin/sh
+
+set -e
+
+MAJOR_MINOR="${1}"
+
+if test -z "${MAJOR_MINOR}"
+then
+	cat <<-EOF >&2
+		This script creates the necessary files for for a new x.y minor release which includes fast, stable and, when appropriate, EUS channel files with required metadata for automation.
+		Usage:
+		  ${0} MAJOR_MINOR
+		For example:
+		  ${0} 4.10
+	EOF
+	exit 1
+fi
+
+CHANNELS="$(printf '%s\n%s' fast stable)"
+
+MAJOR="${MAJOR_MINOR%%.*}"
+MINOR="${MAJOR_MINOR##*.}"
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (changes requested):
@wking There are no mention of this script in the readme or any other documentation in this git repo. Can you add a markdown file for graph-data maintainers and add couple of lines about this script?

wking:
> There are no mention of this script in the readme or any other documentation in this git repo.

I was going to doc in our internal policy repo, because the docs in this repo are not about policy, and these GA release mechanics are about policy.

LalatenduMohanty:
> I was going to doc in our internal policy repo, because the docs in this repo are not about policy, and these GA release mechanics are about policy.

I am not opposed to document this in  internal policy repository. However because the scripts are committed to this repository it is intuitive to search for documents here.

wking:
Of the executables in `hack`:

```console
$ find hack -type f -executable
hack/generate-weekly-report.py
hack/release-ga.sh
hack/show-edges.py
hack/stabilization-changes.py
hack/validate-blocked-edges.py
```

Only one is documented in a separate file:

```console
$ find hack -type f -executable | while read P; do F="$(basename "${P}")"; git --no-pager grep "${F}"; done
hack/release-ga.sh:exec "${DIR}/stabilization-changes.py"
```

Is there some wording I could add to the `Usage:...` message to make the command self-documenting for folks who have discovered it and feel like they want more context?

> However because the scripts are committed to this repository it is intuitive to search for documents here.

I'm optimistic that we can train teammates to hit our internal policy docs first, instead of assuming there will be nothing there and looking in public repos for details of possibly-internal policy.  If we do have local policy docs here, wouldn't that encourage that sort of short-cutting, and risk folks missing some important, internal detail?

PR #1668, opened 2022-03-28
Bumping the minor_min to 4.8.9999 for 4.9 build

LalatenduMohanty:
This PR should be considered after we merge https://github.com/openshift/cincinnati-graph-data/pull/1663. This is done for bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=2068601 

### DIFF
--- a/build-suggestions/4.9.yaml
+++ b/build-suggestions/4.9.yaml
@@ -2,7 +2,7 @@
 # If you specify an arch it must have the full set of values
 ---
 default:
-  minor_min: 4.8.14
+  minor_min: 4.8.9999
   minor_max: 4.8.9999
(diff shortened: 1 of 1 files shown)


sdodson:
I would really prefer we just stay on top of blocking edges as necessary rather than painting ourselves into a corner like this and never being able to change our minds as to whether or not we allow 4.8 to 4.9.future-z until we've fully addressed the issue at hand. This is an evolving situation.

wking:
Seems like we will continue with one-offs like #1698, to give ourselves the flexibility to unblock these update recommendations if we change our mind.  If we keep using that pattern, we won't use this temporary-floor-raising approach.

PR #1686, opened 2022-03-30
channels/stable-4.7: Promote 4.6.56

wking:
It was promoted to the feeder stable by 0000000000 (Version of internal-channels/stable.yaml from internal-channels/stable.yaml, 2022-03-30) 0:00:00.109335 ago.


wking:
Oops, I'll get a better hash.

PR #1720, opened 2022-04-06
hack/release-feature-freeze: Script feature freeze

wking:
So folks can call a script instead of reading docs to prepare a new minor release for feature candidate builds.  Like 288b5daa57 (#1613), but for the earlier lifecycle event.

Also a second commit running this script for 4.11, which generates similar stuff to what we got in 4.10 via #1220 and #1281.


LalatenduMohanty (approved):
/lgtm

PR #1730, opened 2022-04-11
Promote 4.7.47 to fast channel

jottofar:


channels/fast-4.8.yaml
```diff
@@ -47,6 +47,7 @@ versions:
 - 4.7.44
 - 4.7.45
 - 4.7.46
+- 4.7.47
```
LalatenduMohanty, line 50: I do not see 4.7.47 in https://github.com/openshift/cincinnati-graph-data/blob/master/internal-channels/fast.yaml

LalatenduMohanty: We need to add 4.7.47 to the internal fast channel as the errata is public https://access.redhat.com/errata/RHSA-2022:1166

LalatenduMohanty (changes requested):

LalatenduMohanty (approved):
/lgtm

PR #1738, opened 2022-04-12
channels/candidate-4.10: Promote 4.10.9999

wking:
It was promoted to the feeder candidate by ed459352d8 (testing, 2022-04-12) 0:01:36.871226 ago.


wking:
Local testing for stabilization changes.

PR #1806, opened 2022-04-22
4.8.38 and 4.9.30 had misaligned kernel and kernel-rt packages

sdodson:
4.8.38 will get an RHCOS update and become 4.8.39
4.9.30 will be dropped this week


LalatenduMohanty (approved):
/lgtm

PR #1847, opened 2022-05-03
hack/release-stable-minor: Script minor-update stablization

wking:
Continuing in the spirit of 288b5daa57 (#1613) and d45fac659e (#1720).  These lifecycle events aren't frequent, and automating them decreases the odds that we fumble an unfamiliar transition.

I've left off the final:

```console
$ git add channels/*
$ git commit
```

but we could do that in the future if we wanted to control the commit message.

### DIFF
--- a/hack/release-stable-minor.sh
+++ b/hack/release-stable-minor.sh
@@ -1,0 +1,55 @@
+#!/bin/sh
+
+set -e
+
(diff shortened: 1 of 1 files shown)


PratikMahajan (approved):
/lgtm

PR #1911, opened 2022-05-18
Promote 4.9.z to 4.10 stable channel

jottofar:
Replaced the existing filter to include 4.9.z releases and ran the
release-stable-minor script to generate the new content.


wking:
Carried in #1912.

PR #1932, opened 2022-05-20
blocked-edges/4.10.*parallel-ceph_fsync: Handle multiple ceph_health_status series

wking:
Some clusters have a number of these, mostly differing by instance:

```
  ceph_health_status{..., container="mgr", endpoint="http-metrics", instance="10.128.4.12:9283", job="rook-ceph-mgr", namespace="openshift-storage", pod="rook-ceph-mgr-a-6886bc75b8-m6xkd", prometheus="openshift-monitoring/k8s", receive="true", service="rook-ceph-mgr", ...}  1
  ceph_health_status{..., container="mgr", endpoint="http-metrics", instance="10.128.4.16:9283", job="rook-ceph-mgr", namespace="openshift-storage", pod="rook-ceph-mgr-a-6886bc75b8-m6xkd", prometheus="openshift-monitoring/k8s", receive="true", service="rook-ceph-mgr", ...}  1
  ...
```

But we don't care how many there are, we just care if there are any.  Wrapping `ceph_health_status` in a `group(...)` call ignores all the labels and aggregates down to a single series.

This should avoid `EvaluationFailed` issues like:

    invalid PromQL result length must be one, but is 2

### DIFF
--- a/blocked-edges/4.10.10-parallel-ceph_fsync.yaml
+++ b/blocked-edges/4.10.10-parallel-ceph_fsync.yaml
@@ -8,6 +8,6 @@ matchingRules:
 - type: PromQL
   promql:
     promql: |
-      label_replace(ceph_health_status, "ceph", "yes", "", "")
+      label_replace(group(ceph_health_status), "ceph", "yes", "", "")
       or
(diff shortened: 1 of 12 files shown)


LalatenduMohanty (approved):
/lgtm

PR #1965, opened 2022-05-26
blocked-edges/4.10.16: Parallel ceph_fsync risk for 4.10.4 and later

sdodson:
Extend conditional edges into 4.10.16 as we're still waiting on the
kernel fix.


PratikMahajan (approved):
/lgtm

PR #1974, opened 2022-05-31
Do not suggest 4.9.35 in 4.10 EUS

LalatenduMohanty:


wking:
Addressed via #1975.

PR #2059, opened 2022-06-15
hack/stabilization-changes: Gate 4.(y-1).z into fast-4.y, etc. on update paths to 4.y

wking:
We sometimes tombstone or otherwise do not release a 4.y.z in a given week, when we do release 4.(y-1).z'. Hypothetical example:

1. Some bug shows up which makes 4.98.0-to-4.99 conditionally risky, and we've raised `minor_min` to 4.98.25 for new 4.99 builds.
2. Week 1: 4.98.90 and 4.99.50 are built. All is well.
3. Week 2: 4.98.91 and 4.99.51 are built, but 4.99.51 gets tombstoned.
4. Week 3: 4.98.91 enters fast channels, including fast-4.99.
5. Alice is on 4.98.0, and decides to move to 4.99 via `fast-4.99`. There are no recommended updates to 4.99.50 or 4.9.51 because of the long-standing `minor_min` floor.
6. Alice sees recommendations to 4.98.90 and 4.98.91, and heads off to 4.98.91 because it fixes more bugs.
7. Oh no, there are no recommendations to get from 4.98.91 to 4.99. Alice is stuck.
8. Week 4: 4.99.52 enters fast channels, with an update path from 4.98.91, and Alice can finally update.

What we want is to withhold 4.98.91 from `fast-4.99` until it has an update path into 4.99, so when Alice tries to get to 4.99, we send her through 4.98.90, saving her the week+ of waiting for a new 4.99.z release that can update from 4.98.91.

Most of the difficulty here is figuring out whether there will be an update path.  We could teach the script to compute those based on local graph-data information and registry requests like [`hack/show-edges.py` does][1].  But walking Quay tags and pulling and inspecting release images is slow and tedious, and we probably don't need this sort of fanciness for `candidate-...` channels. So this commit sticks with our current behavior for `candidate-...`, uses the script's existing update service request logic to get update recommendations for `candidate-4.y`, and expects those same recommendations to hold true for `fast-4.y` and later if the source and target releases were both in the `fast-4.y` or later channel (which is how Cincinnati works today, and likely to be how it continues to work in the future).

[1]: https://github.com/openshift/cincinnati-graph-data/blob/e27b285dc639aeb01040dde926fc15bbc7659f7a/hack/show-edges.py#L200-L222

### DIFF
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -92,7 +92,7 @@ def stabilize_channel(name, channel, channels, channel_paths, **kwargs):
         feeder_promotion = feeder_promotions[version]
         yield from stabilize_release(
             version=version,
-            channel_name=name,
+            channel=channel,
             channel_path=channel_paths[name],
             delay=delay,
             errata=errata,
@@ -101,7 +101,7 @@ def stabilize_channel(name, channel, channels, channel_paths, **kwargs):
             **kwargs)
 
 
-def stabilize_release(version, channel_name, channel_path, delay, errata, feeder_name, feeder_promotion, cache, waiting_notifications=True, github_token=None, **kwargs):
+def stabilize_release(version, channel, channel_path, delay, errata, feeder_name, feeder_promotion, cache, waiting_notifications=True, github_token=None, **kwargs):
     now = datetime.datetime.now()
     version_delay = now - feeder_promotion['committer-time']
     errata_public = False
@@ -110,7 +110,8 @@ def stabilize_release(version, channel_name, channel_path, delay, errata, feeder
         errata_uri, errata_public = public_errata_uri(version=version, channel=feeder_name, cache=cache)
         if errata_uri:
             public_errata_message = ' {} is{} public.'.format(errata_uri, '' if errata_public else ' not')
-    if (delay is not None and version_delay > delay) or errata_public:
+    concerns_about_updating_out = get_concerns_about_updating_out(version=version, channel=channel, cache=cache) or ''
+    if not concerns_about_updating_out and ((delay is not None and version_delay > delay) or errata_public):
         path_without_extension, _ = os.path.splitext(channel_path)
         subject = '{}: Promote {}'.format(path_without_extension, version)
         body = 'It was promoted to the feeder {} by {} ({}, {}) {} ago.{}'.format(
@@ -124,29 +125,30 @@ def stabilize_release(version, channel_name, channel_path, delay, errata, feeder
         try:
             pull = promote(
                 version=version,
-                channel_name=channel_name,
+                channel_name=channel['name'],
                 channel_path=channel_path,
                 subject=subject,
                 body=body,
 		github_token=github_token,
                 **kwargs)
         except Exception as error:
-            _LOGGER.error('  failed to promote {} to {}: {}'.format(version, channel_name, sanitize(error, github_token=github_token)))
+            _LOGGER.error('  failed to promote {} to {}: {}'.format(version, channel['name'], sanitize(error, github_token=github_token)))
             yield 'FAILED {}. {} {}'.format(subject, body, sanitize(error, github_token=github_token))
         else:
             yield '{}. {} {}'.format(subject, body, pull.html_url)
     else:
-        _LOGGER.info('  waiting: {} ({}){}'.format(version, version_delay, public_errata_message))
+        _LOGGER.info('  waiting: {} ({}){}{}'.format(version, version_delay, public_errata_message, concerns_about_updating_out))
         if waiting_notifications:
-            yield 'Recommend waiting to promote {} to {}; it was promoted the feeder {} by {} ({}, {}, {}){}'.format(
+            yield 'Recommend waiting to promote {} to {}; it was promoted the feeder {} by {} ({}, {}, {}){}{}'.format(
                 version,
-                channel_name,
+                channel['name'],
                 feeder_name,
                 feeder_promotion['hash'][:10],
                 feeder_promotion['summary'],
                 feeder_promotion['committer-time'].date().isoformat(),
                 version_delay,
-                public_errata_message)
+                public_errata_message,
+                concerns_about_updating_out)
 
 
 def get_promotions(path):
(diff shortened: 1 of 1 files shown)


hack/stabilization-changes.py
```diff
@@ -210,6 +212,30 @@ def public_errata_uri(version, cache=None, **kwargs):
...
+        phase, channel_version = channel.rsplit('-', 1)
+    except ValueError:
+        return  # 'fast' and similar version-agnostic channels do not need updating-out concerns
+    if phase == 'candidate':
+        return  # we need ungated candidate-4.y admission to bootstrap using candidate-4.y update recommendations for gating later phases.
+    channel_major_minor = '.'.join(version.split('.', 2)[:2])
+    if release_major_minor == channel_major_minor:
+        return  # we are concerned about getting from 4.(y-1) and earlier into 4.y, not about movement within 4.y.
+    cincinnati_uri, cincinnati_data = get_cincinnati_channel(cache=cache, channel='candidate-{}'.format(channel_major_minor))
+    paths = FIXME # find version in cincinnati_data and look for updates (possibly multi-hop) to 4.y
```
sdodson, line 227: I think as long as `version` has a path to `release_major_minor + 1` we're good. In multi-hop we would've stopped 4.98.91 from going into eus-4.10 because 4.99.50 never made it in because it lacked an upgrade path to 4.100.

sdodson:
Could this also alert us whenever we remove recommendations which cause dead-ends after the fact? I think we should be granted freedom to preserve continuity in upgrades to X.Y even if it means removing some X.Y-n.z temporarily but perhaps only by raising alarms for humans to look into. I view this as different from removing x.y.z from channel-X.Y but maybe I'm overthinking this.  I'm in favor of accepting this, provided you're done and you've tested, then we can revisit closing smaller gaps in the future.

hack/stabilization-changes.py
wking: Rebased and taught about path walking with 65ca645 -> 936de75.  I started in on your single-hop proposal, but realized that even if 4.99.50-to-4.100 was unconditionally recommended when we decided to promote 4.99.50, it may no longer be recommended when we are considering 4.98.91.  And holding 4.98.91 out of a 4.10 channel in that case will make it more likely that folks on 4.98.(z < 91) get a happy path through, instead of heading to 4.98.91, on to 4.99.50, and then being stuck.

Testing with a partial revert of #2245:

```console
$ git revert fd6471d
...
* Recommend waiting to promote 4.7.54 to eus-4.8; it was promoted the feeder stable by 0a4c588e81 (Merge pull request #2237 from openshift-ota-bot/promote-4.7.54-to-stable, 2022-07-20, 4 days, 0:43:33.877052) No unconditional paths from 4.7.54 to later major or minor releases in https://api.openshift.com/api/upgrades_info/v1/graph?channel=candidate-4.8&arch=amd64
...
```

Which looks good to me.

wking:
Testing:

```console
$ git --no-pager log --oneline -1
674e829 (HEAD -> gate-older-minors-on-updates-to-cap-minor-in-candidate-4.y) hack/stabilization-changes: Gate 4.(y-1).z into fast-4.y, etc. on update paths to 4.y
```

Clear out the final z of each 4.y in `eus-4.10` to set up some promotions.  Also remove 4.9.41, because 4.8.46 is old enough to update through 4.9.41 to 4.10.21:

```console
$ sed -i '/- 4[.]\(8[.]46\|9[.]4[12]\|10[.]22\)/d' channels/eus-4.10.yaml 
```

First time through, only 4.10.21 makes the cut (same major.minor as the channel):

```console
$ hack/stabilization-changes.py 2>/dev/null | grep eus-4.10
* channels/eus-4.10: Promote 4.10.22. It was promoted to the feeder stable by 383e64bf0d (Merge pull request #2212 from openshift-ota-bot/promote-4.10.22-to-stable, 2022-07-18) 6 days, 13:51:59.627324 ago. data://no-token-so-no-pull
* Recommend waiting to promote 4.8.46 to eus-4.10; it was promoted the feeder stable by bb27a83d4e (Merge pull request #2190 from openshift-ota-bot/promote-4.8.46-to-stable, 2022-07-13, 11 days, 15:24:29.634141) No unconditional paths from 4.8.46 to 4.10 in https://api.openshift.com/api/upgrades_info/v1/graph?channel=candidate-4.10&arch=amd64 https://api.openshift.com/api/upgrades_info/v1/graph?channel=candidate-4.9&arch=amd64
* channels/eus-4.10: Promote 4.9.41. It was promoted to the feeder stable by eec9b2f74d (Merge pull request #2184 from openshift-ota-bot/promote-4.9.41-to-stable, 2022-07-12) 12 days, 8:37:39.728757 ago. data://no-token-so-no-pull
* Recommend waiting to promote 4.9.42 to eus-4.10; it was promoted the feeder stable by b8295d6cb8 (Merge pull request #2216 from openshift-ota-bot/promote-4.9.42-to-stable, 2022-07-19, 5 days, 16:13:55.755404) No unconditional paths from 4.9.42 to 4.10 in https://api.openshift.com/api/upgrades_info/v1/graph?channel=candidate-4.10&arch=amd64
```

Second time through, now 4.9.42 makes the cut (because 4.10.22 is in) and 4.8.46 makes the cut (because 4.9.41 is in):

```console
$ hack/stabilization-changes.py 2>/dev/null | grep eus-4.10
* channels/eus-4.10: Promote 4.8.46. It was promoted to the feeder stable by bb27a83d4e (Merge pull request #2190 from openshift-ota-bot/promote-4.8.46-to-stable, 2022-07-13) 11 days, 15:24:54.859633 ago. data://no-token-so-no-pull
* channels/eus-4.10: Promote 4.9.42. It was promoted to the feeder stable by b8295d6cb8 (Merge pull request #2216 from openshift-ota-bot/promote-4.9.42-to-stable, 2022-07-19) 5 days, 16:14:20.858847 ago. data://no-token-so-no-pull
```

wking:
>  I think we should be granted freedom to preserve continuity in upgrades to channel-X.Y even if it means removing some X.Y-n.z temporarily but perhaps only by raising alarms for humans to look into.

I don't like removing releases from channels, because then users who updated to that release in that channel will start seeing [`VersionNotFound`][1].  Instead, for update recommendations leaving 4.10 and later, we could use conditional risks tied to the configured channel.  Or just accept the dead-end until we had a replacement path available.  Or we could shift this sort of dancing from the graph-data promotion script into Cincinnati itself (where phased rollouts openshift/enhancements#427 are intended to help reduce `VersionNotFound` issues, and could be extended to cover this sort of dead-end avoidance).  But as you say, we can punt this sort of thing out to future work, without having to sort it in this pull request.

[1]: https://github.com/openshift/cluster-version-operator/blob/3db74ebb012838ffdd776ca65210e4d6d66b7472/pkg/cincinnati/cincinnati.go#L144-L149

LalatenduMohanty:
>  But as you say, we can punt this sort of thing out to future work, without having to sort it in this pull request.

+1

LalatenduMohanty (approved):
/lgtm

PR #2069, opened 2022-06-16
Adding conditional risk ClusterVersionPreconditionCheck for 4.11.0-fc.0

LalatenduMohanty:
Validate it with the `show-edges.py` script i.e. `hack/show-edges.py candidate-4.11`

### DIFF
--- a/blocked-edges/4.11.0-fc.0-cluster-verison-precondition-check.yaml
+++ b/blocked-edges/4.11.0-fc.0-cluster-verison-precondition-check.yaml
@@ -1,0 +1,8 @@
+to: 4.11.0-fc.0
+from: 4.10.14
+url: https://bugzilla.redhat.com/show_bug.cgi?id=2072389#c11
+name: ClusterVersionPreconditionCheck
+message: |-
+  The update from 4.10.14 will not proceed because the cluster version operator will not accept the new target release. We suggest updating to 4.10.15 or later versions (which contains the fix) before updating to 4.11.z
+matchingRules:
+- type: Always


blocked-edges/4.11.0-fc.0-cluster-verison-precondition-check.yaml
```diff
@@ -0,0 +1,8 @@
+to: 4.11.0-fc.0
+from: 4[.]10[.]([7-9]|1[0-4])[+].*
+url: https://bugzilla.redhat.com/show_bug.cgi?id=2072389#c11
+name: ClusterVersionPreconditionCheck
```
wking, line 4:
```console
$ git --no-pager show --oneline --check
a933cfe (HEAD, origin/pr/2069) Adding conditional risk ClusterVersionPreconditionCheck for 4.11.0-fc.0
blocked-edges/4.11.0-fc.0-cluster-verison-precondition-check.yaml:4: trailing whitespace.
+name: ClusterVersionPreconditionCheck 
```

nit: trailing whitespace

blocked-edges/4.11.0-fc.0-cluster-verison-precondition-check.yaml
```diff
@@ -0,0 +1,8 @@
+to: 4.11.0-fc.0
+from: 4[.]10[.]([7-9]|1[0-4])[+].*
+url: https://bugzilla.redhat.com/show_bug.cgi?id=2072389#c11
+name: ClusterVersionPreconditionCheck 
+message: |-
+  The update will not proceed because the cluster version operator will not accept the new target release.
```
sdodson, line 6:
```suggestion
  In clusters updating from releases prior to 4.10.15 the update will not proceed because the cluster version operator will not accept the new target release.
```

Does the risk convey the from and to fields in cluster? Even if so maybe they don't grok our regex, but maybe we don't care. Just consider it.

wking, line 6: Follow-up sentence like:

> We suggest updating to 4.10.15 or later versions (which contains the fix) before updating to 4.11.z.

to include the recommended next step from the impact statement?

wking: The `from` regexp is just between graph-data and Cincinnati; by the time Cincinnati is serving update recommendations to the cluster it's been expanded out into explicit A->B pairs like:

```console
$ curl -s 'https://api.stage.openshift.com/api/upgrades_info/graph?channel=stable-4.10' | jq -r '.conditionalEdges[0]' | head
{
  "edges": [
    {
      "from": "4.10.3",
      "to": "4.10.13"
    },
    {
      "from": "4.10.3",
      "to": "4.10.5"
    },
```

And as it happens, the timing of `minor_min` raises got fc.0 built with 4.10.14 as the only vulnerable release baked in as an fc.0 update source, as seen [here][1] and in:

```console
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.11.0-fc.0-x86_64 | grep Upgrades
  Upgrades: 4.10.14, 4.10.15, 4.10.16, 4.10.17
```

So if we wanted to explain the vulnerable set clearly inline here, maybe:

> 4.10.14 clusters asked to update to 4.11 will not proceed, because... 

?

[1]: https://github.com/openshift/cincinnati-graph-data/pull/2069#issuecomment-1158047394

LalatenduMohanty: If the regular expression is causing confusion then I will change it to `from: 4.10.14`

LalatenduMohanty: Fixed in https://github.com/openshift/cincinnati-graph-data/commit/8467cfc174899c716d46c309b721125b65b8ed88

sdodson: It was less about the regex but more giving hints about what they should do. Though like Trevor mentioned we may want to focus on providing those details in the reference.

Anyway, fine with it as you've updated it.

blocked-edges/4.11.0-fc.0-cluster-verison-precondition-check.yaml
wking: Fixed in 8467cfc174.

PR #2072, opened 2022-06-16
Blocking z stream upgrade paths to 4.9.38 and 4.9.39

LalatenduMohanty:
Because of https://bugzilla.redhat.com/show_bug.cgi?id=2095264#c9

Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

PR #2089, opened 2022-06-21
internal-channels/candidate: Tombstone 4.9.39

wking:
Continuing in the wake of 97c2ee424d (#2072) and 5f23d4e251 (#2080).

I'm personally not clear on why 4.9.39 would be worse off than 4.9.38, and we've shipped 4.9.38.  So folks making install-time decisions might be better off with 4.9.39 (which fixed some 4.9.38 issues, although not this OVN static IP one) than they were with the still-supported 4.9.38.  But the folks making the errata-publication decisions have decided not to ship 4.9.39, so we are tombstoning in this repository to keep up (and to keep from adding 4.9.39 to fast and later channels if the errata URI is recycled for a later release).


LalatenduMohanty (approved):
/lgtm

PR #2118, opened 2022-06-28
blocked-edges/4.10.*: Declare ReleaseDataWithHyphenPrefix risk

wking:
Generated by writing `4.10.14-release-data-with-hyphen-prefix.yaml` and then distributing to the other impacted releases with:

```console
$ for Z in $(seq 15 19); do sed "s/4[.]10[.]14/4.10.${Z}/g" blocked-edges/4.10.14-release-data-with-hyphen-prefix.yaml > "blocked-edges/4.10.${Z}-release-data-with-hyphen-prefix.yaml"; done
```

Because the releases that hash to hyphens [depend on the architecture][1], we could have declared two series of risks, one for `to: 4.10.14+amd64` that looked at born-by-4.6 (which might have touched 4.6.64) and one for `to: 4.10.14+s390x` that looked at born-by-4.9 (which might have touched 4.9.11).  That would have reduced the number of clusters who matched the PromQL, and needed to manually evaluate the risk to assess the impacted updates.  But because the current ClusterVersion history pruning [can remove the originally installed version][2], and 4.6 was a while back, we could have clusters that:

1. Touched 4.6.64 and acquired hyphen-prefixed release data.
2. Accumulated enough subsequent updates to [push 4.6 entries out of the history][2].
3. Are now serving `cluster_version{type="initial"}` with 4.7 or later versions.

To reduce that risk, and because we are interested in exposing customers to the conditional risk interfaces to collect feedback, we are using architecture-agnostic risk declarations and using <= 4.9 initial versions as matching rule broad enough to catch all clusters which might be at risk.  Clusters where the risk matches can evaluate their exposure as discussed in the linked solution, and decide to accept the risk if they determine they are not vulnerable.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2091770#c17
[2]: https://issues.redhat.com/browse/OTA-664


wking:
039085e -> 1418879 moves from `from: .*` to `from: 4[.]10[.].*`.  The risk is for everyone trying to leave 4.10.(14 <= z < 20), not just the ones who got there from 4.10.z.  But I'm only declaring the conditional risk for 4.10.z' -> 4.10.(14 <= z < 20), because only 4.10 cluster-version operators will have the conditional risk support. 4.9.z clusters updating into the impacted releases and then bumping into the bug while trying to leave will have to follow [the recommended mitigation steps][1] to get out.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2091770#c6

blocked-edges/4.10.14-release-data-with-hyphen-prefix.yaml
```diff
@@ -0,0 +1,15 @@
+to: 4.10.14
+from: 4[.]10[.].*
+url: https://access.redhat.com/solutions/6965075
+name: ReleaseDataWithHyphenPrefix
+message: |-
+  Clusters updating out of 4.10.14 may, depending on their past history, fail to initiate the update to any later release.  In order to avoid this trouble, we recommend avoiding updates to this release.  The linked solution includes steps to avoid the issue if you do decide to update to 4.10.14.
```
LalatenduMohanty, line 6: We should add ` workaround and` i.e. `The linked solution includes workaround and steps to avoid the issue if you do decide to update to 4.10.14`

LalatenduMohanty (changes requested):

blocked-edges/4.10.14-release-data-with-hyphen-prefix.yaml
LalatenduMohanty: never mind.

LalatenduMohanty (approved):
/lgtm

wking:
I've pushed 1418879 -> c8511bc to drop the 4.10.19 risk declarations, because we never signed 4.10.19, and declaring risks in updates to an unsigned release [breaks e2e][1]:

    Failed to find signatures for digest sha256:299ce2f9520282173c290b3e237aaa22526ab98ddc67604f2c6fc10e342f5940: [\n    \"Error fetching https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/sha256=299ce2f9520282173c290b3e237aaa22526ab98ddc67604f2c6fc10e342f5940/signature-1 - 404 Not Found\", ...

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-graph-data/2118/pull-ci-openshift-cincinnati-graph-data-master-e2e/1541879405234622464

wking:
https://access.redhat.com/solutions/6965075 has been published.

/hold cancel

PR #2093, opened 2022-06-22
channels: Delay 4.8 promotions into stable-4.9 and eus-4.10

wking:
Continuing in the wake of 97c2ee424d (#2072), 5f23d4e251 (#2080), and f2aef6f664 (#2089).

4.8.43 can update to 4.9.38, but not older 4.9.z:

```console
$ hack/show-edges.py candidate-4.9 | grep '4[.]8[.]43 -.* 4[.]9[.]'
4.8.43 -(blocked: None)-> 4.9.38
4.8.43 -(blocked: None)-> 4.9.39
```

So folks who get into 4.8.43 are stuck without a path to safe 4.9.z until [rhbz#2098099][1] ships with the fix.

4.8.42 can update to 4.9.36 and 4.9.37, so they're fine:

```console
$ hack/show-edges.py candidate-4.9 | grep '4[.]8[.]42 -.* 4[.]9[.]'
4.8.42 -> 4.9.36
4.8.42 -> 4.9.37
4.8.42 -(blocked: None)-> 4.9.38
4.8.42 -(blocked: None)-> 4.9.39
```

4.8.42 already made it into stable channels, but 4.8.43 is currently only through fast.  This commit holds 4.8.43 and later 4.8 out of 4.9 and 4.10 channels, to avoid:

1. User on 4.8.old sets `stable-4.9`.
2. User selects the recommended update to 4.8.43.
3. User cannot update to 4.9 until [rbhz#2098099][1] is fixed.

Instead, we will have:

1. User on 4.8.old sets `stable-4.9`.
2. User selects the recommended update to 4.8.42.
3. User can update to 4.9.36 or 4.9.37 immediately.

We will revert this commit once a fix for [rbhz#2098099][1] reaches stable channels.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2098099

PR #2166, opened 2022-07-07
Revert "channels: Delay 4.8 promotions into stable-4.9 and eus-4.10"

wking:
Reverts openshift/cincinnati-graph-data#2093

4.9.40 has entered stable/EUS channels via #2154, #2155, #2156, so there are now paths out from 4.8.tip into 4.9, and we no longer need to delay these promotions.


LalatenduMohanty (approved):
/lgtm

PR #2255, opened 2022-07-26
internal-channels/candidate: Tombstone 4.9.44

LalatenduMohanty:
4.9.44 is dropped as it needs the fix for https://bugzilla.redhat.com/show_bug.cgi?id=2103982#c16

The fix is already available in the latest 4.9.z nightly.
Without the fix it will cause a regression for clusters upgrading from
4.9.z versions.


wking (approved):
/lgtm

PR #2302, opened 2022-08-09
Create fast and stable feeders

sdodson:


LalatenduMohanty (approved):
/lgtm

channels/fast-4.11.yaml
```diff
@@ -0,0 +1,6 @@
+feeder:
+  delay: PT0H
+  filter: 4[.](10|11)[.][0-9].*
+  name: fast
+name: fast-4.11
+versions:
```
LalatenduMohanty, line 6: We need to define versions as an array : versions: [] 
example : https://github.com/openshift/cincinnati-graph-data/commit/3b5b5436880d84c674b27bdbffffced7863f5f86

channels/stable-4.11.yaml
```diff
@@ -0,0 +1,6 @@
+feeder:
+  delay: PT0H
+  filter: 4[.]11[.][0-9].*
+  name: stable
+name: stable-4.11
+versions:
```
LalatenduMohanty, line 6: Same as above

LalatenduMohanty (changes requested):

LalatenduMohanty:
e2e is failing with below error
```
 Error: Failed to deserialize "/code/channels/stable-4.11.yaml": Message(
    "invalid type: unit value, expected a sequence",
    Some(
        Pos {
            marker: Marker {
                index: 92,
                line: 7,
                col: 0,
            },
            path: "versions",
        },
    ),
)
Caused by:
    Failed to deserialize "/code/channels/fast-4.11.yaml": Message(
        "invalid type: unit value, expected a sequence",
        Some(
            Pos {
                marker: Marker {
                    index: 93,
                    line: 7,
                    col: 0,
                },
                path: "versions",
            }, 
```

LalatenduMohanty (approved):
/lgtm

PR #2361, opened 2022-08-16
internal-channels/candidate: Tombstone 4.10.27

wking:
On AWS clusters in us-iso-east-1 (and other regions that don't have dualstack endpoints), 4.10.27 [fails to install][1], and updates to it, from both 4.10 and 4.9, [fail to complete][1].  We could release 4.10.27 and protect existing clusters by declaring a conditional update risk, but currently tombstoning the release is the only mechanism for protecting folks from installing the broken release.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2115267


LalatenduMohanty (approved):
/lgtm

PR #2364, opened 2022-08-16
blocked-edges: Escape literal periods in 'from'

wking:
We've built up some unescaped values over time, most recently in 8467cfc174 (#2069).  In this commit, I escape them, to reduce the risk of folks copy/pasting and forgetting that `from` is a regular expression.


LalatenduMohanty (approved):
/lgtm

PR #2410, opened 2022-08-29
Block edges to 4.11.2

joepvd:


wking:
Oops, I hadn't realized this was opened when I filed #2450.  But that's landed now, so we can close this one as "accomplished".

PR #2485, opened 2022-09-12
channels/fast-4.11: promote 4.10.30 and 4.10.31

PratikMahajan:
4.10.30 and 4.10.31 were not autopromited because of StaleInsightsRunLevelLabel, StaleSELinuxPolicies


wking (approved):
And `StaleInsightsRunLevelLabel` is the only current risk currently known for updating from these 4.10 to 4.10.4.  `StaleInsightsRunLevelLabel` only impacts a small fraction of 4.10 clusters.  And it's easy for admins to remove the label, after which they are no longer vulnerable.  So there's no concern (at the moment) about clusters updating to 4.10.30 or 4.10.31 and then not having a recommended `fast-4.11` path to 4.11.

/lgtm

PR #2497, opened 2022-09-14
Add 4.10.32 to fast-4.11

sdodson:
Currently the only valid upgrade from 4.10.32 would be 4.11.4 which has a condition hanging off of it so it wouldn't currently be added.


wking:
Carrying on the pattern begun in #2485.

LalatenduMohanty (approved):
/lgtm

PR #2552, opened 2022-09-27
blocked-edges/4.11.6: Declare OVNNetworkPolicyLongName risk

PratikMahajan:
A regression introduced in 4.11.6 may lead to OVN control plane failure and workload disruption.


wking (approved):
/lgtm

PR #2553, opened 2022-09-27
blocked-edges/4.11.6-ovn-namespace: Adding a .yaml suffix

wking:
Fixing a typo from c641333f4a (#2552).  Otherwise the blocked-edges parsers thing this is informative docs, and not an actual risk declaration.  With this change:

```console
$ hack/show-edges.py candidate-4.11 | grep ' 4[.]11[.]6$'
4.10.16 -(blocked: OVNNetworkPolicyLongName)-> 4.11.6
4.10.17 -(blocked: OVNNetworkPolicyLongName)-> 4.11.6
4.10.18 -(blocked: OVNNetworkPolicyLongName)-> 4.11.6
...same OVNNetworkPolicyLongName for everything else going into 4.11.6...
```

which is what we want.

PR #2554, opened 2022-09-27
hack/validate-blocked-edges: Require '.yaml' suffix

wking:
So we catch issues in presubmit CI, to avoid needing post-merge fixups like 4d23a343a9 (#2553).  I've set things up so that we can add `.md` or similar to the allow-list in the future, if we wanted to allow `blocked-edges/README.md`, etc.

### DIFF
--- a/hack/util.py
+++ b/hack/util.py
@@ -7,11 +7,15 @@ import subprocess
 import yaml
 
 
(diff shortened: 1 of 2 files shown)


LalatenduMohanty (approved):
/lgtm

PR #2608, opened 2022-10-06
Populate 4.10.34 and 4.10.35 into fast-4.11

sdodson:
The automation hasn't added these because there are only conditional updates to 4.11.6 and 4.11.7.


sdodson:
Alternatively we agreed that we'd promote with conditional update paths available, @wking is working on that and if it's available soon I'm fine waiting on these edges another day.

LalatenduMohanty:
/hold 

We should check why the bot did not promote this as it should with the change https://github.com/openshift/cincinnati-graph-data/commit/299823467dc55b2c13298a53025e6f114c389ed5

LalatenduMohanty:
Bot created pull requests https://github.com/openshift/cincinnati-graph-data/pull/2610 and https://github.com/openshift/cincinnati-graph-data/pull/2611

PR #2660, opened 2022-10-20
Remove 4.11.10 from candidate

thegreyd:
We're tombstoning 4.11.10
slack thread: https://coreos.slack.com/archives/CJARLA942/p1666275149464379


internal-channels/candidate.yaml
```diff
@@ -620,6 +620,4 @@ versions:
 
 - 4.11.9
 
-- 4.11.10
```
wking, line 623: We don't like removing versions from channels, because then folks who installed this in candidate channels might get `VersionNotFound` issues when looking for updates.  What we want instead is to not promote the release to the `fast` channel, which you can accomplish by leaving this line alone, and adding a 4.11.10 entry and description/bug link up in [the `tombstones` block][1].

[1]: https://github.com/openshift/cincinnati-graph-data/blob/master/internal-channels/candidate.yaml#L2-L81

thegreyd: Makes sense, added now

wking (approved):
/lgtm

PR #2661, opened 2022-10-20
Adding conditional risk for 4.10.38

LalatenduMohanty:

### DIFF
--- a/blocked-edges/4.10.38-node-creation-fails.yaml
+++ b/blocked-edges/4.10.38-node-creation-fails.yaml
@@ -1,0 +1,13 @@
+to: 4.10.38
+from: .*
+url: https://issues.redhat.com/browse/OCPBUGS-2628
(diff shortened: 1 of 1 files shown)


blocked-edges/4.10.38-node-creation-fails.yaml
```diff
@@ -0,0 +1,11 @@
+to: 4.10.38
+from: .*
+url: https://issues.redhat.com/browse/OCPBUGS-2628
+name: NodeCreationFails
+message: |-
+  Ading new nodes to the exsiting cluster will fail.
```
wking, line 6: We should mention something about OpenStack here and in the `name`.  Maybe `OpenStackNodeCreationFails` and:

> OpenStack clusters on 4.10.38 may fail to provision new nodes.

Or maybe it's machines?  Or something...

wking (approved):
/lgtm

PR #2480, opened 2022-09-09
*: Prepare for 4.12 feature freeze

wking:
Like 7203323d38 (#1720), but for 4.12.  Generated with:

```console
$ hack/release-feature-freeze.sh 4.12
```

PR #2758, opened 2022-11-15
*: Prepare for 4.13 development

wking:
Like 2d428f86bc (#2480), but for 4.13.  Run well before feature-freeze, because for 4.13 we plan on cutting sprintly `4.13.0-ec.#` and including them in candidate-4.13.  Generated with:

```console
$ hack/release-open.sh 4.13
$ git add build-suggestions/4.13.yaml channels/*4.13.yaml
```

### DIFF
--- a/build-suggestions/4.13.yaml
+++ b/build-suggestions/4.13.yaml
@@ -1,0 +1,7 @@
+default:
+  minor_min: 4.12.0-rc.0
+  minor_max: 4.12.9999
+  minor_block_list: []
+  z_min: 4.13.0-ec.0
+  z_max: 4.13.9999
+  z_block_list: []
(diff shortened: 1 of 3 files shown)


build-suggestions/4.13.yaml
```diff
@@ -0,0 +1,7 @@
+default:
+  minor_min: 4.12.0
```
sdodson, line 2: I think we need 4.12.0-rc.0 here?

sdodson: For now at least.

wking: Thanks; fixed in 6464eab -> 3eb8e55.

PR #2759, opened 2022-11-16
blocked-edges/4.11.*: Declare ARM64SecCompError524 regression

wking:
Although as I point out in [RUN-1668](https://issues.redhat.com//browse/RUN-1668), [rhbz#2140163][2] makes it sound like the risk might be 4.10-to-4.11, not 4.11-to-4.12, so we may need to refine this later.

[2]: https://bugzilla.redhat.com/show_bug.cgi?id=2140163#c2

### DIFF
--- a/blocked-edges/4.11.0-arm64-seccomp-error-524.yaml
+++ b/blocked-edges/4.11.0-arm64-seccomp-error-524.yaml
@@ -1,0 +1,13 @@
+to: 4.11.0
+from: 4[.]10[.].*
+url:  https://issues.redhat.com/browse/RUN-1668
+name: ARM64SecCompError524
+message: |-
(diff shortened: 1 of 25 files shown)


aleskandro:
Hello @wking, please note that this error doesn't happen during the upgrade from 4.11 to 4.12. It happens on any 4.11 installation and not (only) during upgrades.

wking:
I've pivoted to using 4.10 -> 4.11 as the regression boundary in 596737d -> c35ce29.

andymcc:
I think we should merge this - we're working on a fix but it does mean that a certain container density will prevent bits from working right now. We may be able to come up with a mitigation but we can always adjust the upgrade matrix once we have more info.

PR #2909, opened 2022-12-15
COS-1942: blocked-edges/4.12.*: Declare AWSOldBootImages

wking:
This one is sticky, because we don't have PromQL access to boot-image age, so we cannot automatically distinguish between "born in 4.(<=2) and still uses the old boot images" and "born in 4.(<=2), but has subsequently updated boot images".  And because of a cluster-version operator bug, we don't necessarily have access to the cluster's born-in release anyway.  The CVO bug fix went back via:

* [4.11.0](https://bugzilla.redhat.com/show_bug.cgi?id=2097067#c12)
* [4.10.24](https://bugzilla.redhat.com/show_bug.cgi?id=2108292#c6)
* [4.9.45](https://bugzilla.redhat.com/show_bug.cgi?id=2108619#c6)
* [4.8.47](https://bugzilla.redhat.com/show_bug.cgi?id=2109962#c6)
* [4.7.59](https://bugzilla.redhat.com/show_bug.cgi?id=2117347#c8)
* [4.6.61](https://bugzilla.redhat.com/show_bug.cgi?id=2118489#c6)

So it's possible for someone born in 4.2 to have spend a whole bunch of time in 4.9.z and be reporting a `4.9.0-rc.*` or something as their born-in version.  Work around that by declaring this risk for AWS clusters where the born-in version is 4.9 or earlier, expecting that we'll have this issue fixed soonish, so folks with old boot images will be able to update to a later 4.12, and allowing us to be overly broad/cautious with the risk matching here.

I wrote the 4.12.0-rc.0 content manually, and then copied over to the other 4.12s with:

```console
$ curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=candidate-4.12' | jq -r '.nodes[].version' | grep '^4[.]12[.]' | grep -v '^4[.]12[.]0-rc[.]0$' | while read V; do sed "s/4[.]12[.]0-rc[.]0/${V}/g" blocked-edges/4.12.0-rc.0-AWSOldBootImage.yaml > "blocked-edges/${V}-AWSOldBootImage.yaml"; done
```

### DIFF
--- a/blocked-edges/4.12.0-rc.0-AWSOldBootImage.yaml
+++ b/blocked-edges/4.12.0-rc.0-AWSOldBootImage.yaml
@@ -1,0 +1,21 @@
+to: 4.12.0-rc.0
+from: 4[.]11[.].*
+url: https://issues.redhat.com/browse/COS-1942
+name: AWSOldBootImages
+message: |-
+  4.2 AWS boot images are not compatible with 4.12.0-rc.0, and machines created with them will fail to become nodes.  This risk does not apply if a cluster is not on AWS, was installed as version 4.3 or later, or otherwise uses 4.3 or later boot images.
+matchingRules:
+- type: PromQL
+  promql:
+    promql:
+      topk(1,
+        label_replace(group(cluster_version{type="initial",version=~"4[.][0-9][.].*"}),"born_by_4_9", "yes, so possibly actually born in 4.1 or 4.2", "", "")
+        or
+        label_replace(0 * group(cluster_version{type="initial",version!~"4[.][0-9][.].*"}),"born_by_4_9", "no, born in 4.10 or later", "", "")
+      )
+      * on () group_left (type)
+      (
+        cluster_infrastructure_provider{type="AWS"}
+        or
+        0 * cluster_infrastructure_provider
+      )
--- a/blocked-edges/4.12.0-rc.1-AWSOldBootImage.yaml
+++ b/blocked-edges/4.12.0-rc.1-AWSOldBootImage.yaml
@@ -1,0 +1,21 @@
+to: 4.12.0-rc.1
+from: 4[.]11[.].*
+url: https://issues.redhat.com/browse/COS-1942
+name: AWSOldBootImages
+message: |-
+  4.2 AWS boot images are not compatible with 4.12.0-rc.1, and machines created with them will fail to become nodes.  This risk does not apply if a cluster is not on AWS, was installed as version 4.3 or later, or otherwise uses 4.3 or later boot images.
+matchingRules:
+- type: PromQL
+  promql:
+    promql:
+      topk(1,
+        label_replace(group(cluster_version{type="initial",version=~"4[.][0-9][.].*"}),"born_by_4_9", "yes, so possibly actually born in 4.1 or 4.2", "", "")
+        or
+        label_replace(0 * group(cluster_version{type="initial",version!~"4[.][0-9][.].*"}),"born_by_4_9", "no, born in 4.10 or later", "", "")
+      )
+      * on () group_left (type)
+      (
+        cluster_infrastructure_provider{type="AWS"}
+        or
+        0 * cluster_infrastructure_provider
+      )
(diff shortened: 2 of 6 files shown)


blocked-edges/4.12.0-rc.0-AWSOldBootImage.yaml
```diff
@@ -0,0 +1,21 @@
+to: 4.12.0-rc.0
+from: 4[.]11[.].*
+url: https://issues.redhat.com/browse/COS-1942
+name: AWSOldBootImages
+message: |-
+  4.2 AWS boot images are not compatible with 4.12.0-rc.0, and machines created with them will fail to become nodes.
```
petr-muller, line 6: Can we hint about that it's safe to upgrade if the admin is sure that the cluster is younger than 4.2?

wking: we link to COS-1942, which covers that assessment in the impact statement, but :shrug: I'm fine inlining more text here if you like.  What extra words were you thinking of?

petr-muller: COS-1942 is a wall of text, I'm looking for a fast and easy reassurance for the majority (I'd expect?) case. Let me try to whip up a suggestion.

petr-muller:
```suggestion
  4.2 AWS boot images are not compatible with 4.12.0-rc.0, and machines created with them will fail to become nodes. It is safe to upgrade if a cluster was installed as version 4.3 or later.
```

petr-muller: Not a blocker though, feel free to say that it's not necessary ;)

blocked-edges/4.12.0-rc.0-AWSOldBootImage.yaml
```diff
@@ -0,0 +1,21 @@
...
+        label_replace(group(cluster_version{type="initial",version=~"4[.][0-9][.].*"}),"born_by_4_9", "yes, so possibly actually born in 4.1 or 4.2", "", "")
+        or
+        label_replace(0 * group(cluster_version{type="initial",version!~"4[.][0-9][.].*"}),"born_by_4_9", "no, born in 4.10 or later", "", "")
+      )
+      * on () group_left (type)
+      (
+        cluster_infrastructure_provider{type="AWS"}
+        or
+        0 * cluster_infrastructure_provider
+      )
```
petr-muller, line 21: Checked this on b01 (affected), b02 (not affected, gcp) and b05(not affected, installed as 4.10.9) and the query gives results as expected.

blocked-edges/4.12.0-rc.0-AWSOldBootImage.yaml
wking: I've added:

> This risk does not apply if a cluster is not on AWS, was installed as version 4.3 or later, or otherwise uses 4.3 or later boot images.

which diverges from your suggestion because:

* "It is safe to upgrade" -> "This risk does not apply", because there may be other risks that make updates unsafe.
* Adding a platform qualifier, in case the "AWS boot images" in the first sentence is too subtle for anyone on GCP, vSphere, etc.
* Adding a boot-image mention, because born-on-4.2 isn't a problem if you've subsequently bumped your boot images.

How does that look?

petr-muller: image

petr-muller:
/lgtm
/hold

Not sure whether we want to merge this while COS-1942 is still technically in progress /shrug

wking:
Rebased around #2919 to pick up rc.5 with 1f31ee0 -> 034fa01.

sdodson:
> Not sure whether we want to merge this while [COS-1942](https://issues.redhat.com//browse/COS-1942) is still technically in progress /shrug

I think we should go ahead with this given our understanding of the problem. We can come back and revert if we're wrong. Just my opinion, will leave it up to OTA folks to remove the hold.

PR #2914, opened 2022-12-16
Bump Cincinnati and actix-service

petr-muller:
- Bump Cincinnati and actix-http
- `cargo update --package actix-service`


LalatenduMohanty (approved):
/lgtm

PR #2968, opened 2023-01-06
Convert PromQL risks to Always risks in releases older than 8 weeks

sdodson:
This works around a CVO bug/design decision where we only evaluate one newly enumerated risk every 10 minutes in an effort to avoid overwhelming the monitoring stack with requests. However this creates a bad UX where if there are many risks to evaluate in the set of available update paths it could be N-1 * 10 minutes before the set of recommended updates are computed.

This preserves the notification of issue while largely being a no-op because clusters have, currently, had better update paths for at least 12 weeks.

We intend to fix the CVO bug, but that won't fix the issue in the deployed fleet.

See: https://issues.redhat.com/browse/OCPBUGS-5469


blocked-edges/4.10.14-release-data-with-hyphen-prefix.yaml
```diff
@@ -5,11 +5,4 @@ name: ReleaseDataWithHyphenPrefix
...
 matchingRules:
-- type: PromQL
-  promql:
-    promql: |
-      topk(1,
-        label_replace(group(cluster_version{type="initial",version=~"4[.][0-9][.].*"}),"born_by_4_9", "yes, so possibly touched a hyphen-hashing version", "", "")
-        or
-        label_replace(0 * group(cluster_version{type="initial",version!~"4[.][0-9][.].*"}),"born_by_4_9", "no, so will not have touched a hyphen-hashing version", "", "")
-      )
+- type: type: Always
```
wking, line 8: [validate-blocked-edges][1]:

```
yaml.scanner.ScannerError: mapping values are not allowed here
  in "blocked-edges/4.10.14-release-data-with-hyphen-prefix.yaml", line 8, column 13
```

I expect it wants `type: Always` and not `type: type: Always`.



[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati-graph-data/2968/pull-ci-openshift-cincinnati-graph-data-master-validate-blocked-edges/1611415573508395008#1:build-log.txt%3A53

blocked-edges/4.10.0-fc.2-modified-aws-load-balancer-service.yaml
```diff
@@ -13,9 +13,4 @@ message: |-
 
   This will not cause issues updating between 4.10 releases.  This conditional update is just a demonstration of the conditional update system.
```
wking, line 14: Possibly take this opportunity to just delete this file and `blocked-edges/4.10.0-fc.4-alibaba-storage-fixes.yaml` , since we now have plenty of actual risks, and no longer need these 4.10 feature candidate conditional update demos.

blocked-edges/4.10.11-parallel-ceph_fsync.yaml
```diff
@@ -5,11 +5,4 @@ name: CephParallelFsync
 message: |-
   This update would introduce a CephFS kernel driver regression, exposing a kernel panic when workloads make parallel ceph_fsync calls to the same file.  The update also introduces many bug fixes as described in the errata, so weigh those against the risk of Ceph kernel panics when deciding whether to update or wait for an OpenShift release that also fixes the Ceph regression.
 matchingRules:
-- type: PromQL
-  promql:
-    promql: |
-      topk(1,
-        label_replace(group(ceph_health_status), "ceph", "yes", "", "")
-        or
-        label_replace(0 * group(cluster_version), "ceph", "no", "", "")
```
wking, line 14: possibly shift this PromQL into the `message`? Or the linked bug (although this one links https://bugzilla.redhat.com/show_bug.cgi?id=2076312#c9, which seems to be private)? But the current message is phrased as if we know (or suspect) the cluster is exposed, while with Always this will also trip for clusters we know are not exposed. Or 🤷, maybe that's more polish than we care about for such an old 4.10.z target as these.

sdodson: I'd prefer to just leave it as is. Only 17% of the 4.10 fleet has upgrades to < 4.10.17 and those all have paths to better edges listed more prominently.

blocked-edges/4.10.0-fc.2-modified-aws-load-balancer-service.yaml
sdodson: Should we also clear up the old 4.7 testing conditionals?

wking: Those were real issues, so I'm fine keeping them in place.  But we also have no known clients consuming those conditional risks in 4.6 and 4.7 clusters, so I'm also fine replacing them with an unqualified block into 4.7.4.

wking (approved):
/lgtm

PR #3044, opened 2023-01-23
OTA-858: hack/stabilization-changes: Add risk-extension guards with fixedIn

wking:
With this change, the script will complain like:

```
* Recommend waiting to promote 4.11.24 to fast-4.12; it was promoted the feeder fast by f5a8c3b3ef (Merge pull request #3024 from openshift-ota-bot/promote-4.11.24-to-fast, 2023-01-19, 4 days, 3:10:10.409875)  DuplicateRelease affects 4.11.23.  Either declare a fix version or extend the risk to 4.11.24.  No paths from 4.11.24 to 4.12 in https://api.openshift.com/api/upgrades_info/v1/graph?channel=candidate-4.12&arch=amd64
```

as it wonders if `DuplicateRelease` (which affects updates into 4.11.23) affects 4.11.24 or not.  Graph administrators can either:

* declare `fixedIn` (like I'm doing in this commit) or
* extend the risk to the new release (like we'd done in d44b7a4710 (#3032) and similar.

But the guard should keep us from accidentally promoting releases where we haven't gotten around to evaluating risk extension yet.

### DIFF
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -30,6 +30,7 @@ _GIT_BLAME_COMMIT_REGEXP = re.compile(r'^(?P<hash>[0-9a-f]{40}) .*')
 _GIT_BLAME_HEADER_REGEXP = re.compile(r'^(?P<key>[^ \t]+) (?P<value>.*)$')
 _GIT_BLAME_LINE_REGEXP = re.compile(r'^\t(?P<value>.*)$')
 _GIT_REMOTE_LINE_REGEXP = re.compile(r'^(?P<remote>[^ ]*)\t(?P<uri>(?P<scheme>[^:]*)://(?P<host>[^/]*)/(?P<org>[^/]*)/(?P<repo>[^/.]*)(.git)?) [(](?P<role>.*)[)]$')
+_SEM_VER_REGEXP = re.compile(r'^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$')
 _REMOTE_CACHE = {}
 _SEMANTIC_VERSION_DELIMITERS = re.compile('[.+-]')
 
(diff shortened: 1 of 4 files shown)


petr-muller:
> With this change, the script will complain like:

> * Recommend waiting to promote 4.11.24 to fast-4.12; it was promoted the feeder fast by f5a8c3b3ef (Merge pull request #3024 from openshift-ota-bot/promote-4.11.24-to-fast, 2023-01-19, 4 days, 3:10:10.409875)  DuplicateRelease affects 4.11.23.  Either declare a fix version or extend the risk to 4.11.24.  No paths from 4.11.24 to 4.12 in https://api.openshift.com/api/upgrades_info/v1/graph?channel=candidate-4.12&arch=amd64

The concept is obviously good but I wonder whether we want something more eye-catching in the output so that humans actually act on the warning? I may be wrong but most of the `Recommend waiting to promote 4.8.56 to eus-4.10;...` messages are just churn that do not need an action?

hack/stabilization-changes.py
```diff
@@ -30,6 +30,7 @@
 _GIT_BLAME_HEADER_REGEXP = re.compile(r'^(?P<key>[^ \t]+) (?P<value>.*)$')
 _GIT_BLAME_LINE_REGEXP = re.compile(r'^\t(?P<value>.*)$')
 _GIT_REMOTE_LINE_REGEXP = re.compile(r'^(?P<remote>[^ ]*)\t(?P<uri>(?P<scheme>[^:]*)://(?P<host>[^/]*)/(?P<org>[^/]*)/(?P<repo>[^/.]*)(.git)?) [(](?P<role>.*)[)]$')
+_SEM_VER_REGEXP = re.compile(r'^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$')
```
petr-muller, line 33: :grimacing:  LGTM but I hope it's correct :)

https://regex101.com/r/Tob4oo/1

hack/stabilization-changes.py
```diff
@@ -212,6 +263,46 @@ def public_errata_uri(version, cache=None, **kwargs):
...
+        if key in risks:
+            continue
+        if fixed_in:
+            unfixed.append('{} affects {} and is not fixed until {}.  Extend the risk to {}.'.format(key, previous_version, fixed_in, version))
+        else:
+            unfixed.append('{} affects {}.  Either declare a fix version or extend the risk to {}.'.format(key, previous_version, version))
+
+    if unfixed:
+        return ' '.join(unfixed)
+    return
```
petr-muller, line 303: nit: you can just unconditionally `return ' '.join(unfixed)`:

```console
>>> ' '.join([])
''
```

petr-muller (approved):
Code LGTM with a minor nit -> approving with a hold. The concern about more prominent warning in the output can be discussed / addressed as a followup, we want to guard be in place to protect customer even if we discuss the UX for us.

/lgtm
/hold

hack/stabilization-changes.py
wking: I'll bump my commit message to cite https://semver.org/spec/v2.0.0.html#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string , from which this is copy-pasted

hack/stabilization-changes.py
wking: but I want `None` instead of an empty string, because I like the consistency with the other bare-`return`s.  The consumer is `if concerns_about_risk_extensions`, so your unconditional `join` would work, but it would mean that this function  returns two different values for the different "I don't have any concerns" cases.

hack/stabilization-changes.py
wking: The commit message explains the provenance with 18a5c9b3 -> 0b565b66 .

wking:
> The concept is obviously good but I wonder whether we want something more eye-catching in the output so that humans actually act on the warning?

I've pushed 18a5c9b3 -> 0b565b66, and now output looks like:

    * FAILED ARM64SecCompError524 affects 4.11.25.  Either declare a fix version or extend the risk to 4.11.26. MachineConfigRenderingChurn affects 4.11.25.  Either declare a fix version or extend the risk to 4.11.26.

wking:
And I've pushed 0b565b6 -> 9dbf16d, pointing out that [4.11.26][1] addresses [OCPBUGS-6622](https://issues.redhat.com/browse/OCPBUGS-6622). 

[1]: https://multi.ocp.releases.ci.openshift.org/releasestream/4-stable-multi/release/4.11.26-multi

hack/stabilization-changes.py
petr-muller: :grimacing: I dont review too much Python code recently and was reading this as I would read go. All clear now, sorry for the noise.

petr-muller (approved):
/approve
/hold cancel
/meow

PR #3049, opened 2023-01-24
blocked-edges/4.11.*-MachineConfigRenderingChurn: Link to a KCS

wking:
It turns out that Palash Khaire and Simon Reber had created a Knowledge-Centered Service (KCS) Solution back in November and December, so link that instead of the impact-statement Jira.

The KCS goes into much more detail about mitigation and recovery.  I don't think it explains the race exposure as clearly, but it can always be updated to talk through that more if folks have time.  And the 4.11.z fix should be coming along shortly anyway.


LalatenduMohanty (approved):
/lgtm

PR #3076, opened 2023-01-31
build-suggestions/4.13: Raise minor_min to 4.12.0

wking:
With 4.12 generally available, we no longer need to recomment 4.12.0-pre-release to 4.13.0-pre-release updates.  By not recommending them, we reduce the recommended updates to transitions that users are more likely to be interested in.

Like 7b5f37a767 (#3075), but for the 4.13 build suggestions.


petr-muller (approved):
/lgtm

PR #3111, opened 2023-02-08
hack/stabilization-changes: Optional 'github' package

wking:
When folks do not set a GitHub token, we don't need the `github` package.  Make the package a soft dependency, so it's easier for folks to run the stabilization script locally for testing.


petr-muller (approved):
LGTM

petr-muller:
bah 
/lgtm

LalatenduMohanty:
What happens when PRs get auto `approved` tag? It is still need approved tag from an approver?

PR #3213, opened 2023-03-02
blocked-edges/4.11.*: Declare AWSOldBootImagesLackAfterburn

wking:
Like 034fa0182a (#2909), this one is sticky, because we don't have PromQL access to boot-image age, so we cannot automatically distinguish between "born in 4.1 and still uses the old boot images" and "born in 4.1, but has subsequently updated boot images".  And because of a cluster-version operator bug, we don't necessarily have access to the cluster's born-in release anyway.  The CVO bug fix went back via:

* [4.11.0](https://bugzilla.redhat.com/show_bug.cgi?id=2097067#c12)
* [4.10.24](https://bugzilla.redhat.com/show_bug.cgi?id=2108292#c6)
* [4.9.45](https://bugzilla.redhat.com/show_bug.cgi?id=2108619#c6)
* [4.8.47](https://bugzilla.redhat.com/show_bug.cgi?id=2109962#c6)
* [4.7.59](https://bugzilla.redhat.com/show_bug.cgi?id=2117347#c8)
* [4.6.61](https://bugzilla.redhat.com/show_bug.cgi?id=2118489#c6)

So it's possible for someone born in 4.1 to have spend a whole bunch of time in 4.9.z and be reporting a `4.9.0-rc.*` or something as their born-in version.  Work around that by declaring this risk for AWS clusters where the born-in version is 4.9 or earlier, expecting that we'll have this issue fixed soonish, so folks with old boot images will be able to update to a later 4.12, and allowing us to be overly broad/cautious with the risk matching here.

I wrote the 4.11.0 content manually, and then copied over to the other 4.11s with:

```console
$ curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=candidate-4.11' | jq -r '.nodes[].version' | grep '^4[.]11[.]' | grep -v '^4[.]11[.]0$' | while read V; do sed "s/4[.]11[.]0/${V}/g" blocked-edges/4.11.0-AWSOldBootImageLackAfterburn.yaml > "blocked-edges/${V}-AWSOldBootImageLackAfterburn.yaml"; done
```


petr-muller (approved):
Message and PromQL look good. Holding for ota-monitor to make a decision.

/hold

PR #3250, opened 2023-03-10
validate-blocked-edges: match target version in content and filename

petr-muller:
Prevent gotchas like the one in https://github.com/openshift/cincinnati-graph-data/pull/3243 that needed to be fixed in https://github.com/openshift/cincinnati-graph-data/pull/3249:

When run on https://github.com/openshift/cincinnati-graph-data/pull/3243, this results in the following validation failure:

```console
$ ./hack/validate-blocked-edges.py
  File "/home/pmuller/Projects/RH/cincinnati-graph-data/./hack/validate-blocked-edges.py", line 79, in <module>
    validate_blocked_edges(directory='blocked-edges')
  File "/home/pmuller/Projects/RH/cincinnati-graph-data/./hack/validate-blocked-edges.py", line 13, in validate_blocked_edges
    raise ValueError('invalid blocked edge {}: {}'.format(path, error)) from None
ValueError: invalid blocked edge blocked-edges/4.12.3-leaked-machineconfig.yaml: conditional risk to version 4.12.2 should be declared in a 4.12.2-<...>.yaml file
```

### DIFF
--- a/hack/validate-blocked-edges.py
+++ b/hack/validate-blocked-edges.py
@@ -8,15 +8,20 @@ import util
 def validate_blocked_edges(directory):
     for path, data in util.walk_yaml(directory=directory, allowed_extensions=('.yaml',)):
         try:
-            validate_blocked_edge(data=data)
+            validate_blocked_edge(data=data, path=path)
         except Exception as error:
-            raise ValueError('invalid blocked edge {}: {}'.format(path, error))
+            raise ValueError('invalid blocked edge {}: {}'.format(path, error)) from None
 
 
-def validate_blocked_edge(data):
+def validate_blocked_edge(data, path):
     for prop in ['to', 'from']:
         if prop not in data:
             raise ValueError('{!r} is a required property'.format(prop))
+
+    to = str(data['to'])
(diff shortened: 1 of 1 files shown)


hack/validate-blocked-edges.py
```diff
@@ -8,15 +8,20 @@
 def validate_blocked_edges(directory):
     for path, data in util.walk_yaml(directory=directory, allowed_extensions=('.yaml',)):
         try:
-            validate_blocked_edge(data=data)
+            validate_blocked_edge(data=data, path=path)
         except Exception as error:
-            raise ValueError('invalid blocked edge {}: {}'.format(path, error))
+            raise ValueError('invalid blocked edge {}: {}'.format(path, error)) from None
```
petr-muller, line 13: This is unrelated to the fix, but results in less clutter. It prevents Python from writing out a `During handling of the above exception, another exception occurred` part with a second traceback for the original exception.

petr-muller:
That's some antique Python running there...

petr-muller:
https://github.com/openshift/release/pull/37174

wking (approved):
/lgtm

PR #3327, opened 2023-03-27
SPLAT-984: blocked-edges/4.13.0-*: Declare `NutanixCCMNeedsConfigMap`

petr-muller:
Generated by writing the rc.0 declaration by hand, and then copying out to other 4.13 prereleases with:

```
$ curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=candidate-4.13' | \
    jq -r '.nodes[].version' | \
    grep '^4[.]13[.]' | \
    grep -v '^4[.]13[.]0-rc[.]0$' | while read V;
        do sed "s/4[.]13[.]0-rc[.]0/${V}/g" 4.13.0-rc.0-NutanixCCMNeedsConfigMap.yaml  > "${V}-NutanixCCMNeedsConfigMap.yaml";
    done
```

I am not sure how we eventually limit this, though. This will be resolved by 4.12 having an `Upgradeable=false` in some version (until the user provides the CM). We would bump the minimum version in 4.12 to have this fix, but we would still need to carry the conditional risk for all future 4.13 versions? Because this will never have a `fixedIn` version in 4.13. Not sure.

### DIFF
--- a/blocked-edges/4.13.0-ec.1-NutanixCCMNeedsConfigMap.yaml
+++ b/blocked-edges/4.13.0-ec.1-NutanixCCMNeedsConfigMap.yaml
@@ -1,0 +1,21 @@
+to: 4.13.0-ec.1
+from: 4[.]12[.].*
+url: https://issues.redhat.com/browse/SPLAT-984
+name: NutanixCCMNeedsConfigMap
+message: |-
+  Cloud Controller Manager needs a ConfigMap with user-provided information on Nutanix in OCP 4.13 and upgrades hang when it does not exist.
+matchingRules:
+- type: PromQL
+  promql:
+    promql:
+      topk(1,
+        label_replace(absent(kube_configmap_info{namespace="openshift-cloud-controller-manager", configmap="cloud-conf"}),  "has_cm", "no", "", "")
+        or
+        label_replace(0 * kube_configmap_info{namespace="openshift-cloud-controller-manager", configmap="cloud-conf"},  "has_cm", "yes", "", "")
+      )
+      * on () group_left(type)
+      (
+        cluster_infrastructure_provider{type=~"Nutanix"}
+        or
+        0  * cluster_infrastructure_provider
+      )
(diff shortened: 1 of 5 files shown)


blocked-edges/4.13.0-ec.1-NutanixCCMNeedsConfigMap.yaml
```diff
@@ -0,0 +1,13 @@
+to: 4.13.0-ec.1
+from: 4[.]12[.].*
+url: https://issues.redhat.com/browse/SPLAT-984
```
LalatenduMohanty, line 3: I am not comfortable with linking the impact assessment jira card when the answers are not present. We can use the bug link here till the impact assessment card has more details.

LalatenduMohanty (changes requested):

blocked-edges/4.13.0-ec.1-NutanixCCMNeedsConfigMap.yaml
wking: or we can copy over whatever the relevant bits from the OCPBUGS ticket are?  The impact-statement ticket is mutable, and anything we want to say, we can say there.

sdodson:
/hold
When will we have this in 4.12? I'd prefer we just get that done given the considerations Petr mentions.

sdodson:
/hold cancel
We won't have to declare this indefinitely, only on the limited set of 4.13.0-rc.0 before we land this in 4.12 and raise the 4.12.z bar for upgrades to 4.13.

petr-muller:
/hold

I want to add PromQL to check for the CM presence and also address Lala's comment.

petr-muller:
I have added PromQL that queries the absence/presence of the `cloud-conf` ConfigMap in `openshift-cloud-controller-manager`.

blocked-edges/4.13.0-ec.1-NutanixCCMNeedsConfigMap.yaml
petr-muller: It seems the current consensus is to wait for this one - if we decide we want to merge, we'll want to update SPLAT-984. I'm not doing that for now.

petr-muller:
/close

It is unlikely that we would need this

PR #3357, opened 2023-03-31
blocked-edges: handle cluster_infrastructure_provider zero values

petr-muller:
The `cluster_infrastructure_provider` metric has a zero value when its
type is `None`, so in that case we need to check the metric's presence,
rather than value.

Change prepared with (fish shell, sorry):

```
$ sed -r -i -e 's|(cluster_infrastructure_provider\{.*None.*\})|group(\1)|g' blocked-edges/*.yaml
$ git add blocked-edges
$ for file in (git diff --name-only --staged)
      sed -r -i -e 's|0 \* cluster_infrastructure_provider|0 * group(cluster_infrastructure_provider)|g' $file
  end
```

I have tested the new `OldBootImagesPodmanMissingAuthFlag` PromQL on appci which is a matching AWS cluster, and it works. Hideshi tested the same change on a baremetal cluster (with `"type":"None"`) and it also worked for him.

/cc @wking @LalatenduMohanty

### DIFF
--- a/blocked-edges/4.12.0-OldBootImagesPodmanMissingAuthFlag.yaml
+++ b/blocked-edges/4.12.0-OldBootImagesPodmanMissingAuthFlag.yaml
@@ -15,7 +15,7 @@ matchingRules:
       )
       * on () group_left (type)
       (
-        cluster_infrastructure_provider{type=~"AWS|VSphere|None"}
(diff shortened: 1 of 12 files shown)


wking (approved):
/lgtm

PR #3476, opened 2023-04-13
PatchesOlderRelease: simplify and strengthen the non-recommendation message

petr-muller:
I proposed this change in the original PR but we needed to expedite the blocked edge. I think the detailed messaging is actually not that useful for people, and maybe we do not need to stress out that RH fixes can in theory introduce regressions. I also think that the commitment to eventually deliver a union of all fixes is kinda implicit and we do not need to include it, it just dillutes the 'pls dont upgrade to these versions' core message.


wking (approved):
/lgtm

PR #3509, opened 2023-04-19
OTA-949: README: Define a 1.2.0 filesystem schema for release signatures

wking:
I'm not actually consuming this new structure in Git, so the `version` file remains at 1.1.0.  But we're defining it here because:
    
* We want a versioned schema for passing signature data from `oc-mirror` and other graph-data image producers through to Cincinnati ([for locally run update services](https://issues.redhat.com/browse/OTA-949)),
* We might decide to version-control signatures in the future, and
* We wanted to avoid the overhead of having one versioned filesystem API for Git-to-Cincinnati and another for oc-mirror-graph-data-images-to-Cincinnati.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ And 4.2.14+amd64 would only apply to the amd64 release image.
 
 ### Add Releases To Channels
 
-Edit the appropriate file in `channels/`.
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (approved):
/lgtm

PR #3590, opened 2023-05-09
Retire some older conditional edges which required PromQL eval

sdodson:
This is to work around the series of bugs starting with [OCPBUGS-10514](https://issues.redhat.com/browse/OCPBUGS-10514) where it takes 10mins per unique outstanding `type: PromQL` risk to evaluate in versions prior to 4.12.9, 4.11.34, and 4.10.56. If we reduce the number of outstanding PromQL risks by converting them to Always risks when we've had superseding versions with fixes for those risks we can improve the user experience. FWIW, we've not fixed the throttling to warm in a decent time, we've only updated it so that it evaluates risks on the newest versions first under the assumption that the newest version will have the least conditional updates and we can begin offering updates to the best version as quickly as possible while other, older risks, are being evaluated.

All risks modified here have had fixes in superseding versions for no less than six weeks. This gets us back to one outstanding PromQL risk across all channels.

We've done this previously in https://github.com/openshift/cincinnati-graph-data/pull/2968


LalatenduMohanty (approved):
/lgtm

PR #3622, opened 2023-05-18
*: 4.13.0 is generally available

wking:
Generated by running:

```console
$ hack/release-ga.sh 4.13
```

and then washing through the stabilization script another time or two until we stopped finding promotions:

```console
```


PratikMahajan (approved):
/lgtm

PR #3641, opened 2023-05-23
release-ga.sh: bump z_min to GA version

petr-muller:
Automate https://github.com/openshift/cincinnati-graph-data/pull/3639 / https://github.com/openshift/cincinnati-graph-data/pull/3075 / https://github.com/openshift/cincinnati-graph-data/pull/2369

`sed`ing YAML is a bit ugly but this YAML is simple enough to avoid relying on the user having a pyyaml or something similar installed

### DIFF
--- a/hack/release-ga.sh
+++ b/hack/release-ga.sh
@@ -1,6 +1,11 @@
 #!/bin/sh
 
 set -e
+if ! git diff-index --exit-code --quiet HEAD --
+then
+  echo "This script must be run with clean git state" >&2
+  exit 1
+fi
 
 MAJOR_MINOR="${1}"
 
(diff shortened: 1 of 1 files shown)


hack/release-ga.sh
```diff
@@ -72,6 +72,9 @@ do
 	EOF
 done
 
+# Bump minimal target version to GA one
+sed -i -e "s|z_min: .*|z_min: ${MAJOR}.${MINOR}.0|" "build-suggestions/${MAJOR}.${MINOR}.yaml"
```
wking, line 76: The GA version might not be .0, like 4.10.3:

```console
$ for Y in $(seq 8 13); do grep " 4[.]${Y}[.]" "channels/stable-4.${Y}.yaml" | sort -V | head -n1; done
- 4.8.2
- 4.9.0
- 4.10.3
- 4.11.0
- 4.12.0
- 4.13.0
```

We can probably drop the `exec`  below, run the stabilization script three times or some such, and then take the smallest version out of the new `stable-4.y.yaml` and use that in this injection?  Or maybe that's over-engineering now that we more frequently do GA 4.y.0?  Or maybe even when we don't GA 4.y.0, it's close enough to  be a min version?  Lot of options, and I don't feel strongly about the various trade-offs.

petr-muller: haha I only looked back until 4.11

let me check whether I can easily do something smarter like you suggest. I avoided doing something smarter to avoid dependency on python yaml package, but I havent noticed that we do run stabilization-changes which already use yaml so I can as well use that

petr-muller:
/label tide/merge-method-squash

I have only added this as a separate commit to be able to easily revert

petr-muller:
> Not mergeable. Merge type "squash" disallowed by repo settings 
:thinking: 

/label-remove tide/merge-method-squash

wking (approved):
/lgtm

PR #3650, opened 2023-05-24
Do not warn about risks fixed in earlier candidate to the same channel

petr-muller:
When checking for concerns about risk extensions, take into account also versions that are still candidates for the same channel. It is possible that the (earlier) version that carries the information about the fix (`fixedIn`) is also still just a candidate to the channel, and in that case we would previously log a spurious error message.

For the case where 4.11.40 file contained `fixedIn: 4.11.41` and both versions were still just candidates for stable:

https://github.com/openshift/cincinnati-graph-data/blob/e7567e45e4e1c4857f34f0cf22f341536c1e55af/blocked-edges/4.11.40-KeepalivedMulticastSkew.yaml#L3

Before:
```
* Recommend waiting to promote 4.11.40 to stable; it was promoted the feeder fast by ebf3b7c6fa (Merge pull request #3623 from openshift-ota-bot/promote-4.11.40-to-fast, 2023-05-18, 6 days, 8:35:15.178187)
* FAILED KeepalivedMulticastSkew affects 4.11.39.  Either declare a fix version or extend the risk to 4.11.41.
```

After:
```
* Recommend waiting to promote 4.11.40 to stable; it was promoted the feeder fast by ebf3b7c6fa (Merge pull request #3623 from openshift-ota-bot/promote-4.11.40-to-fast, 2023-05-18, 6 days, 8:38:59.720912)
* Recommend waiting to promote 4.11.41 to stable; it was promoted the feeder fast by 362a228041 (Merge pull request #3644 from openshift-ota-bot/promote-4.11.41-to-fast, 2023-05-24, 12:22:06.721565)

```

### DIFF
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -90,6 +90,11 @@ def sem_ver_less_than(a, b):
     if b_groups['prerelease'] and not a_groups['prerelease']:
         return False
 
+    # Reaching here means releases are equal unless both sides
+    # have a prerelease segment
+    if not (a_groups['prerelease'] or b_groups['prerelease']):
+        return False
+
     # cheating vs. the SemVer spec.
     return a_groups['prerelease'] < b_groups['prerelease']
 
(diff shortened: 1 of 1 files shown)


hack/stabilization-changes.py
```diff
@@ -89,6 +89,8 @@ def sem_ver_less_than(a, b):
         return True
     if b_groups['prerelease'] and not a_groups['prerelease']:
         return False
+    if not (a_groups['prerelease'] or b_groups['prerelease']):
+        return False
```
wking, line 93: What is this case about?  Releases that only differ by `buildmetadata`?

petr-muller: This handles equality when prerelease is not present. We need to handle equality because adding candidates to the list of releases considered includes the candidate currently processed. Without equality the script failed with `KeyError` on the `return` line that indexes with `prerelease`.

I'll add a comment.

wking (approved):
/lgtm

PR #3654, opened 2023-05-24
stabilization-changes.py: code cleanups

petr-muller:
I think this script is now "production" enough to be treated as production code, I'd like to start refactoring it slowly for readability, starting with code cleanups:

- stabilization-changes: fix mixed indentation
- stabilization-changes: populate line number for error output
- stabilization-changes: use set literals
- stabilization-changes: avoid setting some names only conditionally
- stabilization-changes: extract entrypoint to main() to avoid setting globals
- stabilization-changes: use triple quotes for docstrings
- stabilization-changes: remove unncessary parentheses
- stabilization-changes: avoid shadowing `error` names

### DIFF
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -60,7 +60,7 @@ def parse_iso8601_delay(delay):
 
 
 def sem_ver_less_than(a, b):
-    "Returns true if a is less than b, per https://semver.org/spec/v2.0.0.html#spec-item-11"
+    """Returns true if a is less than b, per https://semver.org/spec/v2.0.0.html#spec-item-11"""
     a_match = _SEM_VER_REGEXP.match(a)
     if not a_match:
         raise ValueError('invalid semantic version {!r}'.format(a_match))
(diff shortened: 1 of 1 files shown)


hack/stabilization-changes.py
```diff
@@ -474,10 +474,10 @@ def notify(message, webhook=None):
 
 
 def promote(version, channel_name, channel_path, subject, body, upstream_github_repo, push_github_repo, github_token, upstream_branch, labels=None):
+    upstream_remote = get_remote(repo=upstream_github_repo)
+    branch = 'promote-{}-to-{}'.format(version, channel_name)
```
wking, line 478: Hmm, maybe we should drop `in {}/{}` from the `has already been promoted to` error message below instead?  Because it looks like the check is just "that version is already in that channel", and doesn't inspect to decide how the version got into the channel.  Then we could leave these variables under the `if github_token` branch, because there would be no later consumers.

petr-muller: We could, but  there are `branch` consumers later though. They are protected by the `if not github_token: ... return pull` guard. I dislike semantic code dependencies like that but I guess I can live with that :)

wking (approved):
/lgtm

PR #3672, opened 2023-05-26
OCPCLOUD-2067: Set up `SpuriousCCMOUpgradeableCond` for 4.12.{14-17}

petr-muller:
Based on Scott's analysis and Mike's information I propose setting up a conditional risk for the most affected versions to softly guide clusters to 4.12.18 which should have a significantly lower incidence rate.


blocked-edges/4.12.17-SpuriousCCMOUpgradeableCond.yaml
```diff
@@ -0,0 +1,10 @@
+to: 4.12.17
+from: .*
+fixedIn: 4.12.18
```
petr-muller, line 3: It's not technically fully fixed but the bot would nag us if we did not have this

PR #3785, opened 2023-06-27
hack/exposure-length: Add a script showing known-exposure length

wking:
Timeline for an update-risk bug:

1. Releases named with vulnerable code.
2. Someone notices the issue.
3. Update risk declared.
4. Release with the fix named.
5. `fixedIn` added so that release can enter `candidate-*`.

The duration `T2 - T1` can be long, because many of these risks are rare corner cases.  `T3 - T2` can be short, or it can be long if it takes us some time to understand the severity of the issue.  This script measures `T5 - T3`, so we have a convenient handle on that phase.

### DIFF
--- a/hack/exposure-length.sh
+++ b/hack/exposure-length.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+#
+# Risk declaration until fixed release declared.  Usage:
+#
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (approved):
/lgtm

PR #3786, opened 2023-06-27
OCPNODE-1705: blocked-edges/4.13.*-PerformanceProfilesCPUQuota: Declare new risk

wking:
The PromQL approach is similar to what we used in c641333f4a (#2552), except:

* I'm using the PerformanceProfiles resource.
* I'm using `apiserver_storage_objects` instead of `cluster:usage:resources:sum`, to cut out [one lossy layer of indirection][1].

Generated by writing the 4.13.0 declaration by hand, and then copying out to other 4.13 releases with:

```console
$ curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=candidate-4.13' | jq -r '.nodes[].version' | grep '^4[.]13[.]' | grep -v '^4[.]13[.]0$' | while read V; do sed "s/4[.]13[.]0/${V}/g" blocked-edges/4.13.0-PerformanceProfilesCPUQuota.yaml > "blocked-edges/${V}-PerformanceProfilesCPUQuota.yaml"; done
$ git add blocked-edges/4.13.*PerformanceProfilesCPUQuota.yaml
```

[1]: https://github.com/openshift/cluster-monitoring-operator/blob/70dd5c9a414448ace1d07090730510d5922b6b18/jsonnet/rules.libsonnet#L291-L292C20

### DIFF
--- a/blocked-edges/4.13.0-PerformanceProfilesCPUQuota.yaml
+++ b/blocked-edges/4.13.0-PerformanceProfilesCPUQuota.yaml
@@ -1,0 +1,13 @@
+to: 4.13.0
+from: 4[.]12[.].*
+url: https://issues.redhat.com/browse/OCPNODE-1705
+name: PerformanceProfilesCPUQuota
+message: |-
+  Pods may not execute properly because of CPU quota issues on clusters using PerformanceProfiles.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      group(max_over_time(apiserver_storage_objects{resource="performanceprofiles.performance.openshift.io"}[1h]) > 0)
+      or
+      0 * group(max_over_time(apiserver_storage_objects[1h]))
--- a/blocked-edges/4.13.0-ec.1-PerformanceProfilesCPUQuota.yaml
+++ b/blocked-edges/4.13.0-ec.1-PerformanceProfilesCPUQuota.yaml
@@ -1,0 +1,13 @@
+to: 4.13.0-ec.1
+from: 4[.]12[.].*
+url: https://issues.redhat.com/browse/OCPNODE-1705
+name: PerformanceProfilesCPUQuota
+message: |-
+  Pods may not execute properly because of CPU quota issues on clusters using PerformanceProfiles.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      group(max_over_time(apiserver_storage_objects{resource="performanceprofiles.performance.openshift.io"}[1h]) > 0)
+      or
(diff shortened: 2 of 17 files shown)


wking:
Testing out the PromQL, here's a cluster without any PerformanceProfiles, with the expected 0:

image

And here's simulating the presence of the exposing resource by substituting in `clusterversions.config.openshift.io`, with the expected 1:

image

And here's simulating `apiserver_storage_objects` scraping failing by substituting `apiserver_storage_objectsX`, with the expected `No datapoints found`:

image

blocked-edges/4.13.0-PerformanceProfilesCPUQuota.yaml
```diff
@@ -0,0 +1,13 @@
+to: 4.13.0
+from: 4[.]12[.].*
+url: https://issues.redhat.com/browse/OCPNODE-1705
+name: PerformanceProfilesCPUQuota
+message: |-
+  Clusters using PerformanceProfiles may have CPU quota issues on vulnerable 4.13.
```
petr-muller, line 6: I'd consider being less vague about the observable impact and maybe not mention "vulnerable 4.13" which is kinda implied?

How about something like "Pods may not execute properly because of CPU quota issues on clusters using PerformanceProfiles"?

blocked-edges/4.13.0-PerformanceProfilesCPUQuota.yaml
```diff
@@ -0,0 +1,13 @@
+to: 4.13.0
+from: 4[.]12[.].*
+url: https://issues.redhat.com/browse/OCPNODE-1705
```
petr-muller, line 3: Looks like the jira is not completed yet, we should make sure it is before we merge this.

wking: I've updated the `Description` in the Jira ticket to be an impact statement and moved the ticket to `Code Review`.  We can close once this pull lands.

blocked-edges/4.13.0-PerformanceProfilesCPUQuota.yaml
wking: Works for me; updated in 62262842 -> b5738722.

petr-muller (approved):

petr-muller:
```
openshift-release-dev/ocp-release with layer_digest sha256:a85a707d9f1238d8b214379958ac7de19eb2b5aa631057eb07bdf1b7bdc3eff9: request failed with status 500 Internal Server Error
```

sadge quay

wking:
I'm ok not waiting on Quay.

/override ci/prow/e2e

PR #3792, opened 2023-06-30
Remove KubeletRestartBug risk

sdodson:
While this is a definite regression there are no normally supported operations which trigger a kubelet restart outside of a drain and reboot cycle. Additionally, the risk is simply that pods get restarted which is an expected pod lifecycle event which workloads must tolerate due to any number of conditions such as eviction policies. Therefore we are removing this risk, we hope that a fix for this defect will be available soon.

We recommend continuing to follow this bug for progress https://issues.redhat.com/browse/OCPBUGS-14605


sdodson:
/hold
Just looking to confirm that the current UX due to multiple PromQL evaluations is still confusing. Test cluster coming up now.

sdodson:
Confirmed that a 4.12.20 cluster switching to 4.13 channel doesn't complete evaluating conditional risks for 10 minutes or longer.

wking (approved):
/lgtm

PR #3887, opened 2023-07-24
OTA-1000: RHSB2023-001: Add for 4.12.23+ -> 4.13 edges

petr-muller:
Created the 4.13.0 file manually, then copied the rest via

```console
$ for x in (seq 1 6)
      cp 4.13.0-RHSB2023001.yaml 4.13.$x-RHSB2023001.yaml
      sed -i "s|4.13.0|4.13.$x|" 4.13.$x-RHSB2023001.yaml 
  end
```


petr-muller:
We need to remember to regenerate the PR at that time

wking:
I've opened #3970 floating follow-ups to catch up through 4.13.8 (which should have all the fixes)

PR #3889, opened 2023-07-25
Bump 4.14 minor min version to 4.13.6

sairameshv:
- Following PR includes changes to save the cgroups context in the node config object https://github.com/openshift/machine-config-operator/pull/3793 
- OCP 4.14+ releases default the cgroups version to "v2" 
- The above changes help in restoring the cgroups context during the upgrade of the clusters 
- Hence, it is required for the clusters to upgrade to 4.13.6 before upgrading to 4.14


LalatenduMohanty (approved):
/lgtm

PR #3920, opened 2023-07-31
Move 4.12.26 promotion to stable up by approximate 36hrs

sdodson:
This version clears up the last conditional edges into 4.12


sdodson:
I've reviewed telemetry data to ensure that we have both
- Sufficient accumulation of signal
- No signs of negative signal compared against recent 4.12.z

PR #3983, opened 2023-08-15
`stable-4.13`: Add 4.12 ⟶ 4.13 edges

petr-muller:
Generated with:

```console
$ while ./hack/release-stable-minor.sh 4.13; and ! git diff --quiet ./channels ./internal-channels/
      git add ./channels/ ./internal-channels/
  end
```

First attempt showed a bunch of errors about `OVNCrashOnMigratedDualStack` missing a `fixedIn` record so I added it
in a preceding commit.


LalatenduMohanty (approved):
/lgtm

PR #3984, opened 2023-08-15
NE-1376: Set up `AWSMintModeWithoutCredentials` on 4.13.* ⟶ 4.14.9

petr-muller:
I do not have an affected cluster at hand but I ran the PromQL on an unaffected AWS cluster and it correctly resolved to 0  so at least that side is good :+1: 

All information provided by @wking , I am just the messenger :)

/cc @wking @LalatenduMohanty


wking (approved):
/lgtm

wking:
/retitle NE-1376: Set up AWSMintModeWithoutCredentials on 4.13.* ⟶ 4.13.9

wking (approved):
/lgtm

PR #4059, opened 2023-08-31
validate-blocked-edges.py: enforce CamelCase names

petr-muller:
When testing https://github.com/openshift/cluster-version-operator/pull/961 I have seen CVO choking on incorrect mocked data because we [assign](https://github.com/openshift/cluster-version-operator/blob/53a5d57c2bd3c92c145dac005fab6e4cdcf8f43b/pkg/cvo/availableupdates.go#L363C8-L364C35) the name of the risk to a `condition.Reason` directly:

```
- lastTransitionTime: "2023-08-30T10:19:05Z"
  message: 'ClusterVersion.config.openshift.io "version" is invalid: status.conditionalUpdates[1].conditions[0].reason:
    Invalid value: "OCBBUGS-13308-01": status.conditionalUpdates[1].conditions[0].reason
    in body should match ''^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$'''
  status: "True"
  type: Failing
```

It seems we never actually tripped this in the past but there is no reason why our checks should not protect us from that.

xref: [OCPBUGS-18454](https://issues.redhat.com//browse/OCPBUGS-18454)

### DIFF
--- a/hack/validate-blocked-edges.py
+++ b/hack/validate-blocked-edges.py
@@ -1,8 +1,13 @@
 #!/usr/bin/env python
 
 import os
-
+import re
 import util
+
+# Risk names must be CamelCase to be assignable to condition.Reason
+# https://github.com/openshift/api/blob/8891815aa476232109dccf6c11b8611d209445d9/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1519-L1520C3
+# Note that our regex is stricter, we should never need names more complicated than this.
+NAME_RE = re.compile(r'^[A-Z][A-Za-z0-9_]*$')
 
 
 def validate_blocked_edges(directory):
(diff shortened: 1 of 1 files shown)


hack/validate-blocked-edges.py
```diff
@@ -1,9 +1,15 @@
...
 
 import os
-
+import re
 import util
 
+# Risk names must be CamelCase to be assignable to condition.Reason
+# https://github.com/openshift/api/blob/8891815aa476232109dccf6c11b8611d209445d9/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1519-L1520C3
+# Note that our regex is stricter, we should never need names more complicated than this.
+NAME_RE_RAW = r'[A-Z]([A-Za-z0-9_]*[A-Za-z0-9])?$'
```
wking, line 10: we can probably simplift to `^[A-Z][A-Za-z0-9_]*$`, which:

* Includes a starting anchor `^` for symmetry with the ending anchor `$`.  We don't strictly need `^`, because `match(...)` is `Matches zero or more characters at the beginning of the string.`, but explicitly including it avoids ambiguity.
* Collapse `([A-Za-z0-9_]*[A-Za-z0-9])?` to the simpler and functionally identical `[A-Za-z0-9_]*`.

hack/validate-blocked-edges.py
```diff
@@ -1,9 +1,15 @@
...
+NAME_RE = re.compile(NAME_RE_RAW)
```
wking, line 11: nit: you don't need the separate `NAME_RE_RAW`, because you can use the `pattern` method to get it back from the compiled regexp:

```console
$ python3 -c 'import re; NAME_RE = re.compile(r"[A-Z][A-Za-z0-9_]*$"); print("must match {!r}".format(NAME_RE.pattern))'
must match '[A-Z][A-Za-z0-9_]*$'
```

hack/validate-blocked-edges.py
petr-muller: Good catch! Addressed.

wking (approved):
/lgtm

PR #4178, opened 2023-10-02
Bump the minimum to 4.13.14 to upgrade into 4.14

sdodson:
This picks up two prerequisites

OCPBUGS-18132: Block 4.14 upgrades with admin ack [on vsphere]
OCPBUGS-19351: Keep Profile status.bootcmdline around
the latter prevents dual reboots on upgrades


LalatenduMohanty (approved):
/lgtm

PR #4326, opened 2023-10-31
blocked-edges/4.14.0-ConsoleImplicitlyEnabled: Drop conditional risk from 4.13

wking:
Temporary hack to avoid declaring conditional updates from 4.13.17 and 4.13.18 to 4.14.0:

```console
[
  {
    "url": "https://issues.redhat.com/browse/OTA-1031",
    "name": "ConsoleImplicitlyEnabled",
    "message": "Clusters with the Console capability disabled will have it implicitly enabled by updating to the target release, and once enabled, capabilities cannot be disabled.",
    "matchingRules": [
      {
        "type": "PromQL",
        "promql": {
          "promql": "1 - max(cluster_version_capability{name=\"Console\"})\n"
        }
      }
    ]
  }
]
```

when f0dc7e86d1 (#4301) wanted those silently dropped.  We'll likely need new Cincinnati code to handle this kind of overlap, in case we discover a risk that applies to 4.13.19 to 4.14.0, but for now, drop the overlapping conditional risk.


wking:
Reverts are cheap, and I don't like the narrow, `Console`-linked risk masking the more serious [lack of a mutated-SCC checker][1], so I'll self-LGTM this one.

[1]: https://issues.redhat.com//browse/OCPBUGS-19472

PR #4338, opened 2023-11-01
Adding 4.14.1 to internal-channels/stable.yaml

LalatenduMohanty:
PR #4336 added 4.14.1 to stable-4.14 and eus-4.14 channels. But we forgot adding it to internal-channels/stable.yaml. This caused the automation to complain about it. Hence adding it back retrospectively.


wking (approved):
/lgtm

PR #4377, opened 2023-11-13
SDN-4196: Set `OVNKubeMasterDNPrestop` on OVN clusters installed before 4.11

petr-muller:
Set up risk on updates to 4.12.41+, 4.13.16+ or 4.14+, on OVN clusters
installed before 4.10.

> clusters that have upgraded from 4.10->4.11 will be vulnerable and
> will be affected by this issue when/if they eventually upgrade to
> 4.12.41+, 4.13.16+ or 4.14+
> ...
> - clusters using the OVNKubernetes CNI and that have previously been upgraded from 4.10
>   to 4.11
> - clusters that were initially installed with 4.11 or newer are not affected.

I created the 4.12.41 file manually and copied it around:

```
$ for ocp12 in (seq 42 43)
    sed -e "s/4.12.41/4.12.$ocp12/" blocked-edges/4.12.41-OVNKubeMasterDSPrestop.yaml > blocked-edges/4.12.$ocp12-OVNKubeMasterDSPrestop.yaml
  end
$ for ocp13 in (seq 16 22)
    sed -e "s/4.12.41/4.13.$ocp13/" blocked-edges/4.12.41-OVNKubeMasterDSPrestop.yaml > blocked-edges/4.13.$ocp13-OVNKubeMasterDSPrestop.yaml
  end
$ for ocp14 in (seq 0 2)
    sed -e "s/4.12.41/4.14.$ocp14/" blocked-edges/4.12.41-OVNKubeMasterDSPrestop.yaml > blocked-edges/4.14.$ocp14-OVNKubeMasterDSPrestop.yaml
  end
```

PR #4379, opened 2023-11-13
SDN-4196: blocked-edges/*-OVNKubeMasterDSPrestop: Precise 'from' expressions

wking:
From [SDN-4196](https://issues.redhat.com/browse/SDN-4196):

> clusters that have upgraded from 4.10->4.11 will be vulnerable and will be affected by this issue when/if they eventually upgrade to 4.12.41+, 4.13.16+ or 4.14+

So this commit increases the precision on cfe51de856 (#4377)'s declarations, so updates into the regression like 4.12.40 -> 4.12.41 are labeled with the risk, while updates within post-regression releases like 4.12.41 -> 4.12.42 are not labeled with the risk (because those clusters must have already addressed the issue, or were never exposed in the first place).

Generated with:

```console
$ sed -i 's/^from: .*/from: 4[.](11[.].*|12[.]([0-9]|[1-3][0-9]|40)[+].*)/' blocked-edges/4.12.*-OVNKubeMasterDSPrestop.yaml
$ sed -i 's/^from: .*/from: 4[.](12[.]([0-9]|[1-3][0-9]|40)|13[.]([0-9]|1[0-5]))[+].*/' blocked-edges/4.13.*-OVNKubeMasterDSPrestop.yaml
$ sed -i 's/^from: .*/from: 4[.](13[.]([0-9]|1[0-5])[+].*|14[.]0-.*)/' blocked-edges/4.14.*-OVNKubeMasterDSPrestop.yaml
```

But the floor raise in 17541bdf29 (#4243) moved us past the regression point in 4.13.z -> 4.14 updates, so with the precise regexp, the risk was no longer used for any edges in that minor version bump:

```console
$ hack/show-edges.py candidate-4.14 | grep '4[.]13[.].*OVNKubeMasterDSPrestop.* 4[.]14[.]'
...no hits...
```

So I'm dropping the 4.14 declarations entirely with:

```console
$ rm -f blocked-edges/4.14.*OVNKubeMasterDSPrestop.yaml
```

17541bdf29's `minor_min` advice and 5592c462a9's `PreRelease` risk should be sufficient to keep folks from older 4.13.z from reaching impacted 4.14 without becoming aware of at least some risk.

[1]: https://issues.redhat.com/browse/SDN-4196


blocked-edges/4.12.41-OVNKubeMasterDSPrestop.yaml
```diff
@@ -1,5 +1,5 @@
 to: 4.12.41
-from: .*
+from: 4[.](11[.].*|12[.]([0-9]|[1-3][0-9]|40)[+].*)
```
petr-muller, line 2: I don't understand the literal `+` part (`[+]`) at the end - should that be a `-` instead to cover the RCs and ECs in the `4.12.0-rc.6` form?

wking: [Docs][1]:

> The `[+].*` portion absorbs [the architecture-suffix](https://github.com/openshift/cincinnati-graph-data/tree/b927aa2f6c38ad5e9a8aa195211e9f2c8d7df542#release-names) from the release name that consumers will use for comparisons.

So basically it's making sure the `[0-9]` case doesn't match the first digit of `4.12.43` and similar, by requiring the next character to be the `+` of the arch-suffix, and not a 3 or other digit.

Yes, it will exclude the `-` of RCs and ECs, but all of these are far enough past GA that they shouldn't have direct-from-candidate update paths.

[1]: https://github.com/openshift/cincinnati-graph-data/tree/b927aa2f6c38ad5e9a8aa195211e9f2c8d7df542#block-edges

petr-muller: OK, TIL, I have never seen these before :monocle_face:

petr-muller (approved):
LGTM but holding for Lala

/hold

LalatenduMohanty (approved):
/lgtm

PR #4469, opened 2023-12-01
blocked-edges/4.12.45-OVNKubeMasterDSPrestop: Extend unfixed risk

wking:
[OCPBUGS-24039](https://issues.redhat.com/browse/OCPBUGS-24039) is still `New`, so not fixed in 4.12.z yet.


wking:
I've self-LGTMed this simple change; not worth waiting for reviewers to wake up.

PR #4541, opened 2023-12-20
OCPCLOUD-2409: blocked-edges/4.14*: Declare AzureDefaultVMType

wking:
Generated by writing the 4.14.1 risk by hand, and then running:

```console
$ (curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=candidate-4.14&arch=amd64' | jq -r '.nodes[] | .version' | grep '^4[.]14[.]' | grep -v '^4[.]14[.][01]$') | while read VERSION; do sed "s/4.14.1/${VERSION}/" blocked-edges/4.14.1-AzureDefaultVMType.yaml > "blocked-edges/${VERSION}-AzureDefaultVMType.yaml"; done
```

I also manually added the silent-drop to 4.14.0.  We have almost entirely avoided silent drops since growing the ability to declare conditional risks in 4.10.  But f0dc7e86d1 (#4301) decided to use silent drops for 4.13.17 and 4.13.18 to 4.14.0.  As described in that commit message, there are trade-offs between silent-drops and an Always risk for those updates.  With this commit, we double-down on the silent-drop approach.

Because we're dropping 4.13.19 to 4.14.0 after it has been a supported update for so long, there is a larger risk (than there was for 4.13.17 and 4.13.18 updates) of customers noticing the drop and being confused about where the 4.13.19 to 4.14.0 update went.  That's why we developed the conditional update system in the first place.

That risk is mitigated by the fact that 4.14.0 is fairly old by now, with many subsequent 4.14.z that fix a number of other issues.  So we do not expect there to be much residual interest in 4.13.19 to 4.14.0 updates.

If it turns out that there is enough "where did 4.13.19 to 4.14.0 go?" support load to warrant a pivot, future work could move us to explicitly-declared risk for all of the issues from 4.13.z to 4.14.0, including the original "4.13.19 adds a guard you want first" that f0dc7e86d1 was delivering.  But the price of that pivot is the:

> A silent drop may mean we do not need to support customers who update from 4.13.17 or 18 directly to 4.14.0 and have some mutated SCCs stomped...

trade-off discussed in f0dc7e86d1.

PR #4545, opened 2023-12-21
OCPCLOUD-2409: blocked-edges/*-AzureDefaultVMType: Born in 4.13 ARO all exposed

wking:
Miguel points out that the exposure set [is more complicated][1] than what I'd done in 45eb9ea7c8 (#4541).  It's:

* Azure born in 4.8 or earlier are exposed.  Both ARO (which creates clusters with Hive?) and clusters created via openshift-installer.
* ARO clusters created in 4.13 and earlier are exposed.

Generated by updating the 4.14.1 risk by hand, and then running:

```console
$ curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=candidate-4.14&arch=amd64' | jq -r '.nodes[] | .version' | grep '^4[.]14[.]' | grep -v '^4[.]14[.][01]$' | while read VERSION; do sed "s/4.14.1/${VERSION}/" blocked-edges/4.14.1-AzureDefaultVMType.yaml > "blocked-edges/${VERSION}-AzureDefaultVMType.yaml"; done
```

I break down the new PromQL logic in the commit message.

[1]: https://issues.redhat.com/browse/OCPCLOUD-2409?focusedId=23694976&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23694976

### DIFF
--- a/blocked-edges/4.14.0-ec.0-AzureDefaultVMType.yaml
+++ b/blocked-edges/4.14.0-ec.0-AzureDefaultVMType.yaml
@@ -8,9 +8,17 @@ matchingRules:
   promql:
     promql: |
       topk(1,
-        label_replace(group(cluster_version{_id="",type="initial",version=~"4[.][0-9][.].*"}),"born_by_4_9", "yes, so possibly actually born in 4.8 or earlier", "", "")
+        group by (name) (cluster_operator_conditions{_id="",name="aro"})
         or
-        label_replace(0 * group(cluster_version{_id="",type="initial",version!~"4[.][0-9][.].*"}),"born_by_4_9", "no, born in 4.10 or later", "", "")
+        (
+          0 * group(cluster_operator_conditions{_id=""})
+          + on () group_left (born_by_4_9)
+          topk(1,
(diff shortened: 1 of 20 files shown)


wking:
Screenshots:

Young AWS cluster.  Not exposed, because not Azure:

image

Evaluating on that AWS cluster, with PromQL tweaked to pretend the risk is that same infra platform, because I don't have a young Azure cluster to run the real PromQL on.  Not exposed, because young and not ARO:

image

Evaluating on that AWS cluster, with PromQL tweaked to pretend the risk is "all clusters with `ingress` ClusterOperator conditions" instead of "...with `aro` ClusterOperator conditions", because I don't have a young ARO cluster to run the real PromQL on.  Exposed, because ARO/ingress:

image

Evaluating on that AWS cluster, with `cluster_operator_conditionsBROKEN`, to show how that falls back to no-matches:

image

Evaluating on an old AWS cluster.  No match, because not Azure:

image

Evaluating on that old AWS cluster, with the PromQL pretending the platform exposure is AWS.  Match, because old and infra-match:

image

PR #4575, opened 2024-01-05
OCPCLOUD-2434: blocked-edges/4.14*: Declare AWSECRLegacyCredProvider

wking:
We don't know what PromQL looks like for "I'm using (or not) ECR" looks like, so this will go out to all AWS clusters, not just the exposed ones.  Hopefully the content of the linked URI allows them to determine whether or not they're actually exposed.  If/when we figure out how to tighten the PromQL, we can file a follow-up pull request to adjust here.

Generated the 4.14.0 and 4.14.1 content by hand and copied out with:

```console
$ curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=candidate-4.14&arch=amd64' | jq -r '.nodes[] | .version' | grep '^4[.]14[.]' | grep -v '^4[.]14[.][01]$' | while read VERSION; do sed "s/4.14.1/${VERSION}/" blocked-edges/4.14.1-AWSECRLegacyCredProvider.yaml > "blocked-edges/${VERSION}-AWSECRLegacyCredProvider.yaml"; done
```

### DIFF
--- a/blocked-edges/4.14.0-ec.0-AWSECRLegacyCredProvider.yaml
+++ b/blocked-edges/4.14.0-ec.0-AWSECRLegacyCredProvider.yaml
@@ -1,0 +1,14 @@
+to: 4.14.0-ec.0
+from: 4[.]13[.].*
+url: https://issues.redhat.com/browse/OCPCLOUD-2434
+name: AWSECRLegacyCredProvider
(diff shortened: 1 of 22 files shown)


wking:
Testing the PromQL:

AWS:

image

GCP:

image

Looks good

LalatenduMohanty (approved):
/lgtm

PR #4578, opened 2024-01-05
blocked-edges/4.15.0-rc.0: Declare EarlyAPICertRotation

wking:
We could write PromQL that would try and identify older clusters, but with rc.1 already cut, I've just gone with Always for simplicity, and to avoid folks updating to rc.0 while young, forgetting to update further, and then aging into the early, broken CA rotation.


LalatenduMohanty (approved):
/lgtm

PR #4677, opened 2024-01-25
CCO-522: blocked-edges/4.15.0-rc.0-GCPMintModeRoleAdmin: Declare risk

wking:
The PromQL logic is pretty similar to 4.13.9-AWSMintModeWithoutCredentials.yaml

The exposure is probably somewhere in the 4.15 ECs or RCs, but I'm naively opening with a blanket 4.14 -> 4.15 to get this out the door. We can refine the threshold later as we pin down exactly when in 4.15 the creds started using custom roles.

Generated by manually writing the 4.15.0-rc.0 risk, and copying out to other 4.15 with:

```console
$ curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=candidate-4.15&arch=amd64' | jq -r '.nodes[] | .version' | grep '^4[.]15[.]' | grep -v '^4[.]15[.]0-rc[.]0$' | while read VERSION; do sed "s/4.15.0-rc.0/${VERSION}/" blocked-edges/4.15.0-rc.0-GCPMintModeRoleAdmin.yaml > "blocked-edges/${VERSION}-GCPMintModeRoleAdmin.yaml"; done
```

### DIFF
--- a/blocked-edges/4.15.0-ec.0-GCPMintModeRoleAdmin.yaml
+++ b/blocked-edges/4.15.0-ec.0-GCPMintModeRoleAdmin.yaml
@@ -1,0 +1,21 @@
+to: 4.15.0-ec.0
+from: 4[.]14[.].*
+url: https://issues.redhat.com/browse/CCO-522
+name: GCPMintModeRoleAdmin
+message: |-
+  GCP clusters in Mint mode may need additional permissions to provision 4.15 CredentialsRequests.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      (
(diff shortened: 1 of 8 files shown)


wking:
GCP `credsremoved` cluster says it's not exposed:

image

I dunno how `credsremoved` should be handled, seems like it would be similar to manual-mode, except that you restore your creds pre-update?  I'm ignoring them for now.

GCP `mint` correctly claims exposed:

image

AWS `mint` correctly claims not-exposed:

image

So that all looks good:

/hold cancel

LalatenduMohanty (approved):
/lgtm

PR #3591, opened 2023-05-09
OTA-924: blocked-edges/4.11.*-KeepalivedMulticastSkew: Explicit _id=""

wking:
[In preparation for HyperShift][1], where we expect to have the cluster-version operator resolving update risks via a PromQL engine that stores data from multiple clusters.  Metrics in that data store use an `_id` label to show which cluster the time-series belongs to ([enhancement][2], implementation [here][3] and [here][4]).  This label does not exist in the
`thanos-querier.openshift-monitoring.svc.cluster.local` store for stand-alone, self-hosted clusters, so limiting to `_id=""` is a no-op there.  And future work for HyperShift-hosted cluster-version operators can teach those operators to substitute this placeholder with `_id="theActualClusterID"`, so they match time-series from their cluster, and do not get distracted by time-series from other, sibling hosted clusters.

Generated with:

```console
$ sed -i 's/cluster_infrastructure_provider{/cluster_infrastructure_provider{_id="",/;s/cluster_infrastructure_provider$/cluster_infrastructure_provider{_id=""}/;s/mcd_update_state{/mcd_update_state{_id="",/;s/mcd_update_state)/mcd_update_state{_id=""})/' blocked-edges/4.11.*-KeepalivedMulticastSkew.yaml
```

[1]: https://issues.redhat.com/browse/OTA-924
[2]: https://github.com/openshift/enhancements/blob/ba306451b99da182d26bd0631f3f1e448930316c/enhancements/monitoring/hypershift-monitoring.md#common-labels
[3]: https://github.com/openshift/hypershift/pull/1151/commits/a0be78d4252affff636325c21bc547d6032e3c4a
[4]: https://github.com/openshift/hypershift/pull/1690/commits/44247e9cb86eb0d5f6e44a6a3173844ea8d2ab36

PR #4763, opened 2024-02-09
IR-461: blocked-edges/4.14.*-AzureRegistryImagePreservation: Look for active registry use

wking:
Some clusters do not excercise the internal image registry, and we do not want to bother those admins about this risk.  Look at `imageregistry_request_duration_seconds_count` (which only records registry activity like `BlobStore.Create`) as a sign that the registry is actively used to handle images.  Compare with
`imageregistry_http_request_duration_seconds_count`, which the registry serves even when it is not actively being used to manage images, as a sign that registry metrics are working, to distinguish between "we know the registry is not being used" and "registry metrics collection is broken, and we're blind to registry use".

[In 4.14, the registry component became an optional component][1], but it is not optional in 4.13.  And it's 4.13 clusters evaluating this PromQL because of the `from: 4[.]13[.].*`.  So we don't have to worry about "what if `imageregistry_http_request_duration_seconds_count` is not available?" cases outside of the "registry metrics collection is broken" case.

The registry does not persist these `*_count` values across container restarts, so there is a window of exposure like:

1. Registry activity increments `imageregistry_request_duration_seconds_count` counters.
2. PromQL correctly identifies "this cluster is using its registry".
3. Registry container restarts, zeroing its `*_count` buckets.
4. If _all_ registry containers restart before there is new registry activity, the `imageregistry_request_duration_seconds_count` signal may go away.
5. New registry activity repopulates `imageregistry_request_duration_seconds_count`, return to step 1.

The chances that all registry pods restart at the same time seems low, outside of:

* Updates or other registry-hosting-node reboots, where clusters that actively use the in-cluster registry are likely to have pods consuming images from that registry need re-pulling, which would drive fresh registry traffic and repopulate `imageregistry_request_duration_seconds_count`.
* Registry reconfiguration, which would not drive fresh registry traffic, but which seems like unlikely admin activity in a cluster where the internal-registry is not being used.

I'm also using the `max_over_time` with a 1 hour lookback to try to smear over small durations of "waiting for new registry activity to repopulate `imageregistry_request_duration_seconds_count`".

And now that the PromQL is getting pretty complicated, I'm also adjusting label management so it's easier to see from the result which clauses are matching for a particular cluster, in case we need to debug why we're matching when we don't expect to, or not matching when we do expect to.

I'm also adding `_id=""` placeholders for HyperShift, see 5cb2e93728 (#3591) for more on that.

Generated by editing 4.14.1's by hand and copying out to the others with:

```console
$ ls blocked-edges/*AzureRegistryImagePreservation.yaml | grep -v '/4[.]14[.]1-A' | while read X; do TO="$(grep '^to: ' "${X}")"; sed "s/^to: .*/${TO}/" blocked-edges/4.14.1-AzureRegistryImagePreservation.yaml > "${X}"; done
```

[1]: https://docs.openshift.com/container-platform/4.14/release_notes/ocp-4-14-release-notes.html#ocp-4-14-optional-capabilities

### DIFF
--- a/blocked-edges/4.14.0-ec.0-AzureRegistryImagePreservation.yaml
+++ b/blocked-edges/4.14.0-ec.0-AzureRegistryImagePreservation.yaml
@@ -8,7 +8,15 @@ matchingRules:
   promql:
     promql: |
       topk(1,
-        cluster_infrastructure_provider{_id="",type="Azure"}
+        group by (type) (cluster_infrastructure_provider{_id="",type="Azure"})
         or
-        0 * cluster_infrastructure_provider{_id=""}
+        0 * group by (type) (cluster_infrastructure_provider{_id=""})
+      )
+      * on () group_left (name, operation, method)
+      topk(1,
+        group by (name) (cluster_operator_conditions{_id="",name="aro"})
+        or
+        group by (operation) (max by (operation) (max_over_time(imageregistry_request_duration_seconds_count{_id=""}[1h])))
+        or
+        0 * group by (method) (max by (method) (imageregistry_http_request_duration_seconds_count{_id=""}))
       )
--- a/blocked-edges/4.14.0-ec.1-AzureRegistryImagePreservation.yaml
+++ b/blocked-edges/4.14.0-ec.1-AzureRegistryImagePreservation.yaml
@@ -8,7 +8,15 @@ matchingRules:
   promql:
     promql: |
       topk(1,
-        cluster_infrastructure_provider{_id="",type="Azure"}
+        group by (type) (cluster_infrastructure_provider{_id="",type="Azure"})
         or
-        0 * cluster_infrastructure_provider{_id=""}
+        0 * group by (type) (cluster_infrastructure_provider{_id=""})
+      )
+      * on () group_left (name, operation, method)
+      topk(1,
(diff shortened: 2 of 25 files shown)


wking:
Testing the new PromQL in [4.13.33][1]'s [azure-ovn-upgrade-4.13-micro][2] run in PromeCIeus, it notices some registry use late in the update after the nodes rolled:

image

image


[1]: https://amd64.ocp.releases.ci.openshift.org/releasestream/4-stable/release/4.13.33
[2]: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-ci-4.13-e2e-azure-ovn-upgrade/1755617151344971776

wking:
And testing vs. 4.13.33's [aws-sdn-upgrade-4.13-micro][1] run in PromeCIeus, it sees both "no registry activity" and "detected registry activity", but knows it is not exposed in either case because it is AWS and not Azure:

image


[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.13-e2e-aws-sdn-upgrade/1755617124904079360)

bennerv:
LGTM on ARO:

image

bennerv:
LGTM on aro:
image

PR #4862, opened 2024-02-28
extend AWSCustomDomainNodesNotReady for latest 4.14 versions

PratikMahajan:
/etc/kuberenetes/node.env is overwritten after node reboot.


wking (approved):
/lgtm

PR #4378, opened 2023-11-13
blocked-edges/4.14.0-*-PreRelease: Point out that 4.14 is GA

wking:
Before this commit, folks on 4.13.9 and other older 4.13.z might think they should update to 4.14.0-rc.2 as their largest recommended hop in candidate-4.14:

```console
$ hack/show-edges.py --cincinnati https://api.openshift.com/api/upgrades_info/graph --root-version 4.13.9 candidate-4.14 | grep '^4[.]13[.]9 '
4.13.9 -> 4.13.10
4.13.9 -> 4.13.11
4.13.9 -> 4.13.12
4.13.9 -> 4.13.13
4.13.9 -> 4.13.14
4.13.9 -> 4.13.15
4.13.9 -(risks: OVNKubeMasterDSPrestop)-> 4.13.16
4.13.9 -(risks: OVNKubeMasterDSPrestop)-> 4.13.17
4.13.9 -(risks: OVNKubeMasterDSPrestop)-> 4.13.18
4.13.9 -(risks: OVNKubeMasterDSPrestop)-> 4.13.19
4.13.9 -(risks: OVNKubeMasterDSPrestop)-> 4.13.21
4.13.9 -(risks: OVNKubeMasterDSPrestop)-> 4.13.22
4.13.9 -> 4.14.0-rc.0
4.13.9 -> 4.14.0-rc.1
4.13.9 -(risks: ConsoleImplicitlyEnabled)-> 4.14.0-rc.2
4.13.9 -(risks: ConsoleImplicitlyEnabled, NetworkNodeIdentityWebhookUserConflict)-> 4.14.0-rc.3
4.13.9 -(risks: ConsoleImplicitlyEnabled, NetworkNodeIdentityWebhookUserConflict)-> 4.14.0-rc.4
```

With this commit:

```console
$ hack/show-edges.py --root-version 4.13.9 candidate-4.14 | grep '^4[.]13[.]9 '
4.13.9 -> 4.13.10
4.13.9 -> 4.13.11
4.13.9 -> 4.13.12
4.13.9 -> 4.13.13
4.13.9 -> 4.13.14
4.13.9 -> 4.13.15
4.13.9 -(risks: OVNKubeMasterDSPrestop)-> 4.13.16
4.13.9 -(risks: OVNKubeMasterDSPrestop)-> 4.13.17
4.13.9 -(risks: OVNKubeMasterDSPrestop)-> 4.13.18
[...]

PR #4863, opened 2024-02-29
blocked-edges/4.15.0-*-PreRelease: Declare prerelease risks and automate

wking:
Updating `release-ga.sh` to automate the process from 5592c462a9 (#4378), so we don't need to remember to declare these at GA-time, and running the new script on 4.15 to declare those risks.


PratikMahajan (approved):
/lgtm

PR #4864, opened 2024-02-29
modify EarlyAPICertRotation risk to include promql

PratikMahajan:
All clusters originally installed on OpenShift Container Platform (OCP) version 4.7 or earlier. Clusters installed on 4.8 or later and new 4.15 installs are unaffected.

due to history pruning in the CVO we cannot reliably detect born in versions less than 4.9.
Therefore the conditional update rule will be updated to omit update recommendations in all clusters born in 4.9 or earlier

### DIFF
--- a/blocked-edges/4.15.0-EarlyAPICertRotation.yaml
+++ b/blocked-edges/4.15.0-EarlyAPICertRotation.yaml
@@ -1,7 +1,14 @@
 to: 4.15.0
-from: .*
+from: 4[.]14[.].*
 url: https://issues.redhat.com/browse/API-1687
 name: EarlyAPICertRotation
-message: Clusters older than around one month will trigger an api-int certificate authority rollout, and bugs in that rollout may break kubelet access to the Kubernetes API service.
+message: Clusters born in 4.7 and earlier will trigger an api-int certificate authority rollout, and bugs in that rollout may break kubelet access to the Kubernetes API service.
 matchingRules:
-- type: Always
+  - type: PromQL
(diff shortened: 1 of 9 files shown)


blocked-edges/4.15.0-EarlyAPICertRotation.yaml
```diff
@@ -1,7 +1,14 @@
 to: 4.15.0
-from: .*
+from: 4[.]14[.].*
 url: https://issues.redhat.com/browse/API-1687
 name: EarlyAPICertRotation
-message: Clusters older than around one month will trigger an api-int certificate authority rollout, and bugs in that rollout may break kubelet access to the Kubernetes API service.
+message: Clusters born in 4.6 and earlier will trigger an api-int certificate authority rollout, and bugs in that rollout may break kubelet access to the Kubernetes API service.
```
sdodson, line 5: Some of the discussion indicated that we believe some early 4.7.z are also vulnerable. Can we change this to just say 4.7 and earlier? No need to be precise about 4.7.z given we're only able to write promql that applies to 4.9 and earlier.

PratikMahajan: done

wking (approved):
/lgtm

PratikMahajan:
image

image

Ran the metrics on old and new cluster to confirm that it works

wking:
/retitle API-1687: modify EarlyAPICertRotation risk to include promql

PR #4892, opened 2024-03-07
blocked versions due to OCPBUGS-30256

pepedocs:
blocked versions due to OCPBUGS-30256

Issue link: https://issues.redhat.com/browse/OCPBUGS-30256

### DIFF
--- a/blocked-edges/4.13.0-AROBrokenAcceleratedNetworking.yaml
+++ b/blocked-edges/4.13.0-AROBrokenAcceleratedNetworking.yaml
@@ -1,0 +1,15 @@
+to: 4.13.0
+from: .*
+url: https://issues.redhat.com/browse/OCPBUGS-30256
+name: AROBrokenAcceleratedNetworking
+message: |-
+  Adding a new worker node will fail for clusters running on ARO with Azure Accelerated Networking.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      (
+        group(cluster_operator_conditions{name="aro"})
+        or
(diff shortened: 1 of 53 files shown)


LalatenduMohanty (changes requested):
We do not need any conditional risk for a cluster which has already updated to 4.13 , as the risk does not change between 4.13 to 4.14 or 4.14 to 4.15 if the cluster already on that version which has the regression. 
So please remove the risk from 4.14.z, 4.15.z versions

blocked-edges/4.13.0-AROBrokenAcceleratedNetworking.yaml
```diff
@@ -0,0 +1,15 @@
+to: 4.13.0
+from: .*
+url: https://issues.redhat.com/browse/OCPBUGS-30256
```
LalatenduMohanty, line 3: Update this link to https://issues.redhat.com/browse/OPNET-479 in all the files included in this PR

LalatenduMohanty (changes requested):

LalatenduMohanty:
>Disagree. The issue that necessitates the edge occurs when a node reboots, as it would during upgrade. We want to minimize instances of this as SRE has to manually remediate each node.

As per the remediation steps we need to delete .link files to fix the issue. So if a cluster has update to 4.13 and the files are deleted, will these files again get created when they update to 4.14.z?

cblecker:
/close
in favour of https://github.com/openshift/cincinnati-graph-data/pull/4903

PR #4966, opened 2024-03-22
Enable 4.11.59 in candidate channel

wking:
Not sure why the ART bot didn't open this pull after creating this branch, but catching it up manually.

PR #4967, opened 2024-03-22
Enable 4.14.18 in candidate channel

wking:
More catching up with branches the ART bot opened but did not create pulls for.


wking:
Rebased around #4966.

wking:
And self-LGTMed, since it's also just lightly-touched, bot-generated content.

PR #4968, opened 2024-03-22
Enable 4.13.38 in candidate channel

wking:
And more catching up for the ART bot that created branches but did not open pulls this week.


wking:
Rebased around #4966 and #4967

wking:
And self-LGTMed again, because I'm just passing along the ART bot's suggestion.

PR #4982, opened 2024-03-26
blocked-edges/4.15.5-EarlyAPICertRotation: Not fixed in 4.15.5

wking:
[OCPBUGS-31384](https://issues.redhat.com/browse/OCPBUGS-31384) is still `New`.


wking:
Self-LGTM, because everyone else has clocked out by now.

PR #5051, opened 2024-04-05
MCO-1091: blocked-edge: set up `CSRNotApprovedBadCerts` on 4.16.0-ec.4

petr-muller:
Limit the exposed clusters to ones installed as 4.13 or lower, as described in [MCO-1091](https://issues.redhat.com//browse/MCO-1091).


LalatenduMohanty (approved):
/lgtm

PR #4958, opened 2024-03-20
Need fixes for OCPBUGS-29290 in 4.14.16 to upgrade to 4.15 safely

sdodson:
ref: https://issues.redhat.com/browse/OCPBUGS-29290?focusedId=24338647&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24338647

PR #5052, opened 2024-04-05
 MCO-1031: Address `AWSCustomDomainNodesNotReady` on 4.14->4.15 paths

petr-muller:
@sdodson [said](https://issues.redhat.com/browse/OCPBUGS-29290?focusedId=24338647&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24338647) :

>> If the customer upgrades through 4.14.16 or later before going into 4.15 then the issue will not present itself. For new 4.15 or new 4.14 clusters the issue also does not manifest

> Therefore we should introduce conditional update risks on to stop recommending updates to all 4.15.z from 4.14.14 - 4.14.15. 

This was previously proposed in https://github.com/openshift/cincinnati-graph-data/pull/4958 but closed, but I think that it is still possible that someone goes to the exposed 4.14, performs the workaround to heal their cluster, then gets bitten *again* when going to 4.15 without going through 4.14.16.


petr-muller:
We tombstoned 4.15.7 and e2e tests failed on fetching its signatures

sdodson:
Needs similar treatment for 4.15.1 and 4.15.4. I wonder if it's just best to add those to the tombstone list?

PR #5091, opened 2024-04-11
Extending the risk to 4.15.9 and adding fixedIn for 4.15.8

LalatenduMohanty:
As the bugfix of OCPBUGS-31807 is not in a 4.15.z version

### DIFF
--- a/blocked-edges/4.15.8-AWSCustomDomainNodesNotReady.yaml
+++ b/blocked-edges/4.15.8-AWSCustomDomainNodesNotReady.yaml
@@ -1,5 +1,6 @@
 to: 4.15.8
 from: 4[.]14[.](14|15)
+fixedIn: 4.15.9
 url: https://issues.redhat.com/browse/MCO-1031
 name: AWSCustomDomainNodesNotReady
 message: |-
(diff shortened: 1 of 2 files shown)


blocked-edges/4.15.8-AWSCustomDomainNodesNotReady.yaml
```diff
@@ -1,5 +1,6 @@
 to: 4.15.8
 from: 4[.]14[.](14|15)
+fixedIn: 4.15.8
```
wking, line 3: Do you mean `fixedIn: 4.15.9`, since 4.15.9 is the first release with only `4.14.(>=16)` update sources?

petr-muller: Yeah, this should be 4.15.9 
/hold

LalatenduMohanty: Fixed it in https://github.com/openshift/cincinnati-graph-data/commit/035b1090f75e489e65d7f1fabba248f6f7f66baa

wking: /hold cancel

wking (approved):
/lgtm

PR #5094, opened 2024-04-11
OPRUN-3267: Adding OLMOperatorsInFailedState risk

LalatenduMohanty:
Both OCPBUGS-31080 and OCPBUGS-24009 has same symptoms and mostlikely related. We are adding a single risk for both of these bugs. However if we come to know if these bugs are separate then we will add separate risks.

Used below script to generate the files

$ curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=candidate-4.15&arch=amd64' | jq -r '.nodes[] | .version' | grep '^4[.]15[.]' | grep -v '^4[.]15[.]0' | grep -v '0-[er]c[.]'| while read VERSION; do sed "s/4.15.0/${VERSION}/" blocked-edges/4.15.0-OLMOperatorsInFailedState.yaml > "blocked-edges/${VERSION}-OLMOperatorsInFailedState.yaml"; done

### DIFF
--- a/blocked-edges/4.15.0-OLMOperatorsInFailedState.yaml
+++ b/blocked-edges/4.15.0-OLMOperatorsInFailedState.yaml
@@ -1,0 +1,15 @@
+to: 4.15.0
+from: 4[.]14[.]*
+url: https://issues.redhat.com/browse/OPRUN-3267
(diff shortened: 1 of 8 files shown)


wking (approved):
/lgtm

PR #5120, opened 2024-04-17
blocked-edges/4.15.8-AzureRegistryImageMigrationUserProvisioned: Fixed in 4.15.9

wking:
The buggy Secret handling is still present in 4.15.9, but it is present in all 4.15.z so far, so no regressions for 4.15-to-4.15 updates.  And 4.14.(>=15) are exposed, so both sides of 4.14.(>=15) to 4.15 are exposed, and no regressions there either.  That leaves 4.14.(<15) to 4.15, and while we have those up through 4.14.8, 4.14.9 was built after 37826e2d89 (#5052) merged, and there are no 4.14.(<15) to 4.15.(>=9) edges to talk about.


PratikMahajan (approved):
/lgtm

PR #5129, opened 2024-04-19
hack/stabilization-changes: Restore waiting_notifications in stabilize_version

wking:
0b160190cb (#5128) moved `waiting_notifications` from `kwargs` into an explicitly handled argument in `stabilize_channel`.  That means we need to explicitly pass it down into `stabilize_version` to keep those version notifications in the daily batch message.


wking:
self-LGTM, because it's late, and I've already dragged everyone else through enough rounds of trying to get this patch-update-connectivity warning business working 😅

PR #5219, opened 2024-05-08
NE-1689: Declare `IngressDegradedOnRouterReloads` for <4.14.23

petr-muller:
OTA consensus was to declare this risk once a fixed version in present in the stable channel, which was expedited for 4.14.23 and it went to stable via https://github.com/openshift/cincinnati-graph-data/pull/5214

I was thinking of maybe making this a conditional risk based on the count of Routes in the cluster, but without knowing a good threshold I would be just speculating. We have a fixed 4.14.23, lets steer folks towards it.


wking (approved):
/lgtm

PR #4339, opened 2023-11-01
blocked-edges/4.14.0-ConsoleImplicitlyEnabled: Drop conditional risk from 4.13 again

wking:
Repeating 897f57f57e (#4326).  I'd thought the tuned 'from' regular expressions in 361e132baa
(#4337) would avoid the issue, but apparently not:

```console
$ curl -s 'https://api.openshift.com/api/upgrades_info/graph?arch=amd64&channel=fast-4.14' | jq -r '.nodes[] | select(.version == "4.14.0").metadata["io.openshift.upgrades.graph.previous.remove_regex"]'
4[.](13[.]19|14[.]0-(ec[.].*|rc[.][01]))
```

showing Cincinnati using the `from` from the `ConsoleImplicitlyEnabled`, and apparently forgetting about the silent-drop from `4.14.0.yaml`:

```console
$ hack/show-edges.py --cincinnati https://api.openshift.com/api/upgrades_info/graph --root-version 4.13.17 fast-4.14
4.13.17 -> 4.13.18
4.13.17 -> 4.13.19
4.13.17 -> 4.14.0  # this unqualified recommendation misses the mutated-SCC guard from 4.14.0.yaml
4.13.18 -> 4.13.19
4.13.18 -> 4.14.0  # this unqualified recommendation misses the mutated-SCC guard from 4.14.0.yaml
4.13.19 -(risks: ConsoleImplicitlyEnabled)-> 4.14.0  # we'd like to keep this
4.13.19 -(risks: ConsoleImplicitlyEnabled)-> 4.14.1
4.14.0 -> 4.14.1
```

This commit is expected to transition us to:

```console
$ hack/show-edges.py --root-version 4.13.17 fast-4.14
4.13.17 -> 4.13.18
4.13.17 -> 4.13.19
4.13.17 -(SILENT-BLOCK)-> 4.14.0  # better than the current unqualified recommendation
4.13.18 -> 4.13.19
4.13.18 -(SILENT-BLOCK)-> 4.14.0  # better than the current unqualified recommendation
4.13.19 -> 4.14.0  # we'd like this to mention ConsoleImplicitlyEnabled, but I'm back to not knowing how to get that to happen
4.13.19 -(risks: ConsoleImplicitlyEnabled)-> 4.14.1
4.14.0 -> 4.14.1
```

PR #5222, opened 2024-05-08
Backfill 4.14.0 risks from 4.14.1 and remove 4.14.0 drop

petr-muller:
Generated with https://github.com/petr-muller/ota/tree/main/cmd/graph-extend-or-fix by basically extending backwards:
```console
for risk in AWSCustomDomainNodesNotReady AWSECRLegacyCredProvider AzureDefaultVMType AzureRegistryImagePreservation ConsoleImplicitlyEnabled IngressDegradedOnRouterReloads
    go run ./cmd/graph-extend-or-fix/... --jira-endpoint=https://issues.redhat.com/ --jira-bearer-token-file=$HOME/Temporary/jira-upgradeblocker-token --graph-repository-path=/home/pmuller/Projects/RH/github.com/openshift/cincinnati-graph-data --risk=$risk --last=4.14.1 --new=4.14.0 --do=extend
end
```

I have not included the formatting changes in AWSCustomDomainNodesNotReady and IngressDegradedOnRouterReloads that already existed on 4.14.0

Initially we wanted to silently drop the 4.14.0 release from the graph but because of [OTA-1043](https://issues.redhat.com/browse/OTA-1043) it is hard to keep it that way, because we usually automate adding new 4.14 risks over individual patch updates and forget to avoid adding one for 4.14.0. This way we struggle to keep the drop in effect (see https://github.com/openshift/cincinnati-graph-data/pull/4339 and https://github.com/openshift/cincinnati-graph-data/pull/4301), so we decided to give up and just bury 4.14.0 in conditional risks.

### DIFF
--- a/blocked-edges/4.14.0-AWSECRLegacyCredProvider.yaml
+++ b/blocked-edges/4.14.0-AWSECRLegacyCredProvider.yaml
@@ -1,0 +1,20 @@
+to: 4.14.0
+from: 4[.]13[.].*
+url: https://issues.redhat.com/browse/OCPCLOUD-2434
+name: AWSECRLegacyCredProvider
+message: AWS clusters that use AmazonEC2ContainerRegistryReadOnly node policies to access ECR are unable to pull images from ECR after updating to exposed 4.14.z.
+matchingRules:
+    - type: PromQL
+      promql:
+        promql: |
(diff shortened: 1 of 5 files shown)


wking (approved):
/lgtm

PR #5226, opened 2024-05-09
AUTH-517: Declare `ServiceAccountsOAuth` for cluster with disabled registry

petr-muller:
~~PromQL works as expected on enabled capability, disabled capability, and defensively, on clusters that are somehow not aware a capability:~~

Trevor suggested we pivot to checking the presence of scaled-up image registry deployment:

cluster runs 2 replicas of registry
first is "no such deployment" -> 1
second is "scaled beyond threshold" -> 0
third is "scaled below threshold" -> 1

For production PromQL we just check the threshold is 0, not 2

image


petr-muller:
/hold

Needs testing, and there is also chance that the impact is lower (tokens that already existed before update maybe continue to work, just ones created post-update would be affected)

wking (approved):

PR #4903, opened 2024-03-08
Add risks for OPNET-479

cblecker:

PR #5289, opened 2024-05-23
OTA-1179: README: Clarify managed-cluster risk scoping

wking:
Point at the fleet-scoped OCPBUGS / `UpgradeBlocker` assessment process, because we don't want managed to declare risks without having that all-fleet assessment (even if the all-fleet assessment happens after a managed-specific risk declaration).  If an all-fleet risk is declared, any managed-specific risk in that space can be removed.  If an all-fleet risk is not declared, service-delivery graph-data admins can make their own call about what to do with a managed-specific risk.

Also point out the possibility of customer-managed admins being confused by managed-specific risks.  For example, ca5795c373 (#4903) was an ARO-specific declaration, but the message said "clusters running on Azure with Accelerated Networking" which nominally includes non-ARO Azure clusters.  Some of those cluster admins asked about why the risk wasn't matching their clusters, and it's because [we saw no evidence of non-ARO clusters being bit by the race][1].  The guidelines I'm adding will hopefully reduce the chance of future managed-specific risk declarations causing similar customer-managed admin confusion.

I'm also adding `_id=""` to the queries as a pattern to support HyperShift and other systems that could query the cluster's data out of a PromQL engine that stored data for multiple clusters.  More context in 5cb2e93728 (#3591).

And I'm also adding ARO-specific PromQL.  Maybe someday ARO will join the other managed flavors in using sre:telemetry:managed_labels, but they haven't yet.

[1]: https://issues.redhat.com/browse/OPNET-479?focusedId=24366396&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24366396


LalatenduMohanty (approved):
/lgtm

PR #5377, opened 2024-06-11
API-1813: blocked-edges/4.16.*-FlowControlV1Beta3Migration: Declare new risk

wking:


wking:
Testing on an old cluster:

image

And a younger cluster:

image

So those look good.

PratikMahajan (approved):
/lgtm

PR #5418, opened 2024-06-18
README: Explain scoping vs. OCP policy and support

wking:
To reduce the chances that an OpenShift-cluster administrator reading the README jumps from "this API allows the graph-data admins to do things like dropping update edges" (which is true) to "so maybe tomorrow my 4.15.0 to 4.15.2 recommendation will be blocked in stable-4.15!" (which is not true).  That won't happen because [4.15 docs][1] say:

> Beginning in OpenShift Container Platform 4.10, when update risks are confirmed, they are declared as Conditional Update risks for the relevant updates.

which is a policy commitment constraining update service changes. Hopefully the new wording adds some clarity on the space between "how the tools could be used" and "how Red Hat will use the tools".

[1]: https://docs.openshift.com/container-platform/4.15/updating/understanding_updates/understanding-update-channels-release.html#conditional-updates-overview_understanding-update-channels-releases

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ This repository manages the Cincinnati graph for OpenShift.
 
 All of this data feeds the OpenShift Update Service, and describes the schema and APIs that graph-data administrators can use to configure that service.
 Nothing in this repository sets policy on which updates are supported for which clusters (which is downstream of the update service).
-Nothing in this repository sets policy for how graph-data administorators decide to use the available graph-data (that policy is internal, but the public commitments are covered in product docs like [these][openshift-channel-docs].
+Nothing in this repository sets policy for how graph-data administrators decide to use the available graph-data (that policy is internal, but the public commitments are covered in product docs like [these][openshift-channel-docs].
 
 ## Workflow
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,6 +3,10 @@
 [Cincinnati][] is an update protocol designed to facilitate automatic updates.
 This repository manages the Cincinnati graph for OpenShift.
 
+All of this data feeds the OpenShift Update Service, and describes the schema and APIs that graph-data administrators can use to configure that service.
+Nothing in this repository sets policy on which updates are supported for which clusters (which is downstream of the update service).
+Nothing in this repository sets policy for how graph-data administorators decide to use the available graph-data (that policy is internal, but the public commitments are covered in product docs like [these][openshift-channel-docs].
```
petr-muller, line 8:
```suggestion
Nothing in this repository sets policy for how graph-data administrators decide to use the available graph-data (that policy is internal, but the public commitments are covered in product docs like [these][openshift-channel-docs].
```

petr-muller (approved):

PR #5457, opened 2024-06-25
SDN-5055: blocked-edges/4.16.0-rc.6-OVNMulticastPacketLoss: Fixed in rc.9

wking:
Stacking on #5456, in case that speeds up CI.

[4.16.0-rc.9][1] contains [OCPBUGS-35835][2].

rc.7 and rc.8 were never pushed into channels, as we keep spinning out new releases hoping each will have the last GA-blocker :).

[1]: https://amd64.ocp.releases.ci.openshift.org/releasestream/4-stable/release/4.16.0-rc.9?from=4.16.0-rc.6
[2]: https://issues.redhat.com/browse/OCPBUGS-35835


wking:
Late for everyone else, so self-LGTM

PR #5490, opened 2024-06-27
GA 4.16

petr-muller:
- **Force release 4.16 into fast**
- **`./hack/release-ga.sh 4.16`**


wking:
CI seems to have gotten stuck, but I'm very confident in this change not breaking anything:

/override ci/prow/e2e
/override ci/prow/images

PR #5566, opened 2024-07-17
blocked-edges/4.13.25+-ARODNSWrongBootSequence: Declare risk

jaitaiwan:
https://issues.redhat.com/browse/OCPBUGS-35300 remains an issue for clusters after 4.13.25.

### DIFF
--- a/blocked-edges/4.13.25-ARODNSWrongBootSequence.yaml
+++ b/blocked-edges/4.13.25-ARODNSWrongBootSequence.yaml
@@ -1,0 +1,13 @@
+to: 4.13.25
+from: 4[.]12[.].*
+url: https://issues.redhat.com/browse/OCPBUGS-35300
+name: ARODNSWrongBootSequence
+message: |-
+  Disconnected ARO clusters or clusters with a UDR 0.0.0.0/0 route definition that are blocking the ARO ACR and quay, are not be able to add or replace nodes after an upgrade
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+        group(cluster_operator_conditions{name="aro"})
+        or
+        0 * group(cluster_operator_conditions)
--- a/blocked-edges/4.13.26-ARODNSWrongBootSequence.yaml
+++ b/blocked-edges/4.13.26-ARODNSWrongBootSequence.yaml
@@ -1,0 +1,13 @@
+to: 4.13.26
+from: 4[.]12[.].*
+url: https://issues.redhat.com/browse/OCPBUGS-35300
+name: ARODNSWrongBootSequence
+message: |-
+  Disconnected ARO clusters or clusters with a UDR 0.0.0.0/0 route definition that are blocking the ARO ACR and quay, are not be able to add or replace nodes after an upgrade
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+        group(cluster_operator_conditions{name="aro"})
+        or
+        0 * group(cluster_operator_conditions)
--- a/blocked-edges/4.13.27-ARODNSWrongBootSequence.yaml
+++ b/blocked-edges/4.13.27-ARODNSWrongBootSequence.yaml
@@ -1,0 +1,13 @@
+to: 4.13.27
+from: 4[.]12[.].*
+url: https://issues.redhat.com/browse/OCPBUGS-35300
+name: ARODNSWrongBootSequence
+message: |-
+  Disconnected ARO clusters or clusters with a UDR 0.0.0.0/0 route definition that are blocking the ARO ACR and quay, are not be able to add or replace nodes after an upgrade
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+        group(cluster_operator_conditions{name="aro"})
+        or
+        0 * group(cluster_operator_conditions)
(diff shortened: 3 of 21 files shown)


blocked-edges/4.13.25-ARODNSWrongBootSequence.yaml
```diff
@@ -0,0 +1,13 @@
+to: 4.13.25
+from: 4[.]12[.].*
```
wking, line 2: This won't include 4.13.(z<25) -> 4.13.25, and you probably want to include those too, right?

wking: from: `4[.](12|13)[.].*` would work for this -> 4.13.25 file, but you'll want something more nuanced in 4.13.26 and later to avoid declaring the risk on 4.13.25 -> 4.13.26 (where the update between two exposed releases isn't regressing).

wking: Something like:

```yaml
from: 4[.](12[.][0-9]*|13[.](1?[0-9]|2[0-4]))[+].*
```

might do it, with the terminal `[+].*` explained [here][1].

[1]: https://github.com/openshift/cincinnati-graph-data/blame/cb30b80c03c2f86bc8d6df8e123673996991d115/README.md#L137-L144

wking: Confirming in a checkout with an alternation to one file, although you'll probably have to wait a long time for the script to fill your local cache with enough 4.y.z metadata to render:

```console
$ git --no-pager log -1 --oneline
5b4b2137 (HEAD, origin/pr/5566) blocked-edges/4.13.25+-ARODNSWrongBootSequence: Declare risk
$ git --no-pager diff
diff --git a/blocked-edges/4.13.26-ARODNSWrongBootSequence.yaml b/blocked-edges/4.13.26-ARODNSWrongBootSequence.yaml
index 2ad51c4a..6c99fbaa 100644
--- a/blocked-edges/4.13.26-ARODNSWrongBootSequence.yaml
+++ b/blocked-edges/4.13.26-ARODNSWrongBootSequence.yaml
@@ -1,5 +1,5 @@
 to: 4.13.26
-from: 4[.]12[.].*
+from: 4[.](12[.][0-9]*|13[.](1?[0-9]|2[0-4]))[+].*
 url: https://issues.redhat.com/browse/OCPBUGS-35300
 name: ARODNSWrongBootSequence
 message: |-
 $ hack/show-edges.py candidate-4.13 | grep ' 4[.]13[.]26$' | sort -V
4.12.16 -(risks: AROBrokenDNSMasq, ARODNSWrongBootSequence, AcceleratedNetworkingRace, OVNKubeMasterDSPrestop)-> 4.13.26
4.12.17 -(risks: AROBrokenDNSMasq, ARODNSWrongBootSequence, AcceleratedNetworkingRace, OVNKubeMasterDSPrestop)-> 4.13.26
...
4.12.45 -(risks: AROBrokenDNSMasq, ARODNSWrongBootSequence, AcceleratedNetworkingRace)-> 4.13.26
4.13.0 -(risks: AROBrokenDNSMasq, ARODNSWrongBootSequence, OVNKubeMasterDSPrestop)-> 4.13.26
...
4.13.24 -(risks: AROBrokenDNSMasq, ARODNSWrongBootSequence)-> 4.13.26
4.13.25 -(risks: AROBrokenDNSMasq)-> 4.13.26
```

hlipsig: Discussed in slack and determined that's enough of a corner case that we can leave for now and re-evaluate as necessary.

jaitaiwan: I think `4[.](12[.].*|13[.]([0-3][0-9]|4[0-5]|[0-9]))[-].*` will do the job if @hlipsig decides to do a further PR. @wking is it supposed to be `[+].*` or `[-].*`?

jaitaiwan: Based on tests here: https://regexr.com/83ed9

wking: you don't need to worry about matching `4.13.45-rc.9`, because #3639 raised `z_min` for new 4.13.z to 4.13.0, so release like 4.13.26 don't bake in those prereleases as possible update sources:

```console
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.13.26-x86_64 | grep Upgrades:
  Upgrades: 4.12.16, 4.12.17, ..., 4.12.45, 4.13.0, 4.13.1, ... 4.13.24, 4.13.25
```

You do need to handle `[+].*` if you want to anchor the tail of the 4.y.z version, because, [as discussed in the README][1], that matches [architecture suffixes][2] like you have in `4.2.14+amd64` (which is how Cincinnati represents the amd64 4.2.14 release internally).  Without the `[+].*` tail anchor, a regular expression like `4[.]13[.][0-9]` would match 4.13.4 (like you want) but it would also match 4.13.40 (which is already exposed to this bug, so we don't want to declare 4.13.40 -> 4.13.41 as a regression risk).

[1]: https://github.com/openshift/cincinnati-graph-data/blame/cb30b80c03c2f86bc8d6df8e123673996991d115/README.md#L137-L144
[2]: https://github.com/openshift/cincinnati-graph-data/blob/cb30b80c03c2f86bc8d6df8e123673996991d115/README.md#release-names

jaitaiwan: Thanks for the explanation that helps

PR #5539, opened 2024-07-10
MCO-1212: Revert "blocked-edges/4.13.7+:OldBootImagesMissingOSReleaseRHELVersion

PratikMahajan:
This reverts commit 412db5fe7d94a2f31166ddf316e62cce933e8629.

reverting this block till we get more clarity on the issue

From the QE testing, we have learnt that older node bootimage scale-up works fine on 4.13 and 4.14 without needing any further fix.
https://issues.redhat.com/browse/MCO-1212?focusedId=25101566&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25101566

PR #5528, opened 2024-07-08
blocked-edges/4.13.7+: Declare OldBootImagesMissingOSReleaseRHELVersion 

PratikMahajan:
Machine boot images from 4.1 and 4.2 are not compatible with some 4.13 and later
OpenShift releases, and machines created with them will fail to become nodes.  This
risk does not apply if a cluster was installed as version 4.3 or later, or otherwise
uses 4.3 or later boot images.

clusters originally installed with OCP 4.1 and 4.2 and upgraded to 4.13.7+,
4.14 or later version are impacted.

Not declaring risks for 4.13.x -> 4.14.x because of z floor bumping.

Created 4.13.7 manually and copied risk for other edges
```shell 
curl -s 'https://api.openshift.com/api/upgrades_info/graph?channel=fast-4.13' | jq -r '.nodes[].version' | grep '^4[.]13[.]' | grep -v '^4[.]13[.][0-7]$' | while read V; do sed "s/4[.]13[.]7/${V}/g" blocked-edges/4.13.7-OldBootImagesMissingOSReleaseRHELVersion.yaml > "blocked-edges/${V}-OldBootImagesMissingOSReleaseRHELVersion.yaml"; done
```

PR #5579, opened 2024-07-18
MCO-1212: blocked-edges/4.16.{0,1,2}: Re-declare `OldBootImagesMissingOSReleaseRHELVersion`

petr-muller:
Originally declared for 4.13.7+ in https://github.com/openshift/cincinnati-graph-data/pull/5528 and then reverted in https://github.com/openshift/cincinnati-graph-data/pull/5539 because we discovered the versions are not actually affected. We confirmed the exposure for 4.16.* and the fix went into 4.16.3 ([OCPBUGS-36330](https://issues.redhat.com/browse/OCPBUGS-36330)) so declaring just for the three earlier versions.

Copied one file from the original #5528, manually adjusted for 4.16.0, then copied with

```
for patch in 1 2
    cp blocked-edges/4.16.0-OldBootImagesMissingOSReleaseRHELVersion.yaml blocked-edges/4.16.$patch-OldBootImagesMissingOSReleaseRHELVersion.yaml
    sed -i "s|to: 4.16.0|to: 4.16.$patch|" blocked-edges/4.16.$patch-OldBootImagesMissingOSReleaseRHELVersion.yaml
end
```

Finally added a `fixedIn: 4.16.3` to the 4.16.2 file.

### DIFF
--- a/blocked-edges/4.16.0-OldBootImagesMissingOSReleaseRHELVersion.yaml
+++ b/blocked-edges/4.16.0-OldBootImagesMissingOSReleaseRHELVersion.yaml
@@ -1,0 +1,12 @@
+to: 4.16.0
+from: 4[.]15[.]*
+url: https://issues.redhat.com/browse/MCO-1212
+name: OldBootImagesMissingOSReleaseRHELVersion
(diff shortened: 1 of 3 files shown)


wking (approved):
/lgtm

PR #5587, opened 2024-07-19
blocked-edges: fix typos in regexes

petr-muller:
Trevor noticed we sometimes have `4[.]X[.]*` instead of `4[.]X[.].*`. The former would capture multiple dots where our intent is to capture a single dot followed by anything (usually patch numbers). It happens to work because we match on substrings, but fixed regexes express our intent better and are more readable for everyone.

Edited with:

```
sed -E -i -e 's|4\[\.\]([0-9]*)\[\.\]\*|4[.]\1[.].*|' blocked-edges/*
```


wking (approved):
/lgtm

PR #5588, opened 2024-07-19
SDN-5145: Set up `OVNIPsecPausedMCPConnectivity` for 4.14 -> 4.15 paths

petr-muller:
OVN clusters with enabled IPsec with multiple worker MachineConfigPools may lose SDN connectivity during updates when at least one worker MCP is paused.


petr-muller:
/hold

PromQL needs to be reviewed and tested (and the blocked edge spread for all 4.15 releases).

blocked-edges/4.15.0-OVNIPsecPausedMCPConnectivity.yaml
```diff
@@ -0,0 +1,20 @@
...
+    promql: |
+      (
+        group by (ipsec) (label_replace(max_over_time(ovnkube_master_ipsec_enabled{_id=""}[1h]), "ipsec", "enabled", "", "") == 1)
+        or on (_id)
+        0 * group by (ipsec) (label_replace(max_over_time(ovnkube_master_ipsec_enabled{_id=""}[1h]), "ipsec", "disabled", "", "") == 0)
+      ) and on (_id) (
+        group by (resource) (max_over_time(apiserver_storage_objects{_id="",resource="egressips.k8s.ovn.org"}[1h]))
+        and on (_id)
+        group by (resource) (max_over_time(apiserver_storage_objects{_id="",resource="machineconfigpools.machineconfiguration.openshift.io"}[1h] ) > 2)
+      )
```
petr-muller, line 18: "uses OVN" (likely redundant with "ipsec enabled" but IMO does not hurt
and
"has more than two MachineConfigPools"

It seems we cannot check pausedness so we overabstract and make false positives but with IPsec guard that may be acceptable (we do not falsely warn all clusters with custom MCPs but only ones with IPsec)

petr-muller:
4.14.0-0.nightly-2024-07-18-144042 OVN clustrer right after install

The metric is really called `ovnkube_controller_ipsec_enabled`, not `ovnkube_master_ipsec_enabled`. OVN is detected enabled correctly, IPsec disabled correctly, over 2 MCPs correctly false, entire query results in 0: not matching. Correct.

image

petr-muller:
Enabled IPsec following https://docs.openshift.com/container-platform/4.14/networking/ovn_kubernetes_network_provider/configuring-ipsec-ovn.html

IPsec is now detected as enabled, everything else is the same, whole query still reports 0 (because we do no have more MCPs)

image

petr-muller:
Created an MCP. After a moment (metrics need to propagate) the MCP qeury started returing a result and the whole query returns a 1 value = cluster matches. Looks good.

image

wking (approved):
/lgtm

PR #5683, opened 2024-08-08
Block edges for ARO clusters with systemd dependency loops

lranjbar:
Related to https://issues.redhat.com/browse/OCPBUGS-37534

Investigation is still ongoing to determine full impact for born in versions. Currently, ARO SRE is seeing 100% failure rate into versions with this code change listed above.


bennerv:
ARO Cluster born on 4.13+ not evaluating :+1: 
image


ARO cluster born on <4.13 evaluating :+1: 
image

PR #5693, opened 2024-08-09
Update AROSystemDLooping promql to have no errors

lranjbar:
Fixes small error seen in the promql in #5683 

Cluster born in 4.12 matching
image


bennerv:
/lgtm

Fixes the error that was hitting before, but was still successfully evaluating.  Thank you @lranjbar

PR #5705, opened 2024-08-12
NO-JIRA: Tombstone 4.15.26

Davoska:
The RHBA advisory link was attached as opposed to the RHSA. The version was repromoted.

See: https://redhat-internal.slack.com/archives/CJARLA942/p1723218481456569


sdodson:
/override ci/prow/e2e
This only prevents release from moving forward we can skip e2e

PR #5721, opened 2024-08-14
Update AROSystemDLooping risk based off testing

lranjbar:
After much testing we've determined that simply dodging the issue in 4.13.z will avoid this bug. Removing edge blocks for 4.14.z, 4.15.z and 4.16.z.


bennerv:
/lgtm 

KCS article should have context on why the edges were pulled.

wking:
Quay 502 is orthogonal to this change:

/override ci/prow/e2e

PR #5735, opened 2024-08-15
Fixes shipped

sdodson:
AzureSystemDLooping fixed in 4.13.48
SRIOVFailedToConfigureVF fixed in 4.14.35, 4.15.28, 4.16.8

Require 4.16.8 so that fixes for OCPBUGS-38182 are present before upgrading to 4.17.


PratikMahajan (approved):
/lgtm

PR #5845, opened 2024-09-06
Clarify SRIOVFailedToConfigureVF message

sdodson:


petr-muller (approved):
/hold

LGTM. Clarification works for me (I would not be able to act on the previous one, I understand the new one better), holding for the case this needs an actual SR-IOV SME to approve. Unhold as needed.

petr-muller:
/hold cancel

:shrug: I think the new description is better, we can always followup if needed

PR #6223, opened 2024-11-08
SDN-5477: blocked-edges/4.14.40-OVNlibreswan: Declare new risk

wking:


blocked-edges/4.14.40-OVNlibreswan.yaml
```diff
@@ -0,0 +1,12 @@
+to: 4.14.40
+from: .*
+url: https://issues.redhat.com/browse/SDN-5477
+name: OVNlibreswan
+message: OVN clusters with IPsec enabled will regress on some libreswan CVEs affecting the ovnk container.
```
sdodson, line 5:
```suggestion
message: OVN clusters with IPsec enabled will regress on some libreswan CVEs.
```
I'm pretty sure it was accurate as is, but there's a smearing of when libreswan is used from RHCOS or from the container. For instance there are tickets outstanding that mean to pin RHCOS as well and it's not clear to me if those are moving forward, stopped because of this regression, or even necessary. This detail doesn't seem critical to understanding whether or not they want to accept the risk so drop it.

sdodson (changes requested):

PR #6268, opened 2024-11-19
OTA-1392: channels/stable-4.16: Add 4.14.z

wking:
Folks on `4.(y-2).z` may wish to go to `4.y` but not qualify for the [Extended Update Support][1] or [Extended Update Support Term 2][2] life cycle phases.  By adding `4.(y-2).z` to `stable-4.y`, we can pivot [control-plane-only docs][3] to use `stable-4.16` and similar in their examples, to make it more clear that that paused-MachineConfigPool update process is orthogonal to the EUS-ness of the life cycle phase or channel name.

[1]: https://access.redhat.com/support/policy/updates/openshift#eus
[2]: https://access.redhat.com/support/policy/updates/openshift#eust2
[3]: https://docs.openshift.com/container-platform/4.16/updating/updating_a_cluster/control-plane-only-update.html


wking:
/hold

Discussing this internally, it should require more than our usual amount of consensus before merging, because once we add 4.y.z releases to channels, we cannot remove them without triggering `VersionNotFound` issues in any 4.y.z clusters who have selected the newly-available-to-them channel.

channels/stable-4.16.yaml
```diff
@@ -1,9 +1,49 @@
 feeder:
   delay: PT0H
-  filter: 4[.](15|16)[.][0-9].*
+  filter: 4[.](14|15|16)[.][0-9].*
   name: stable
 name: stable-4.16
 versions:
+- 4.14.0
```
wking, line 7: Excluding `fast-4.16` here is [causing][1]:

```
Error: Error in channel ordering: ["Release 4.14.30 present in stable-4.16, but not in fast-4.16...
```

Probably easier to do the same thing to `fast-4.16` and `candidate-4.16`, but I'll leave those out for now, while we wait to collect feedback, in case folks think it's worth just doing `stable-4.16` (with a CI `/override ...`?) for some reason.

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cincinnati-graph-data/6268/pull-ci-openshift-cincinnati-graph-data-master-e2e/1859293525741408256#1:build-log.txt%3A25

sdodson: Yeah I've assumed we'd just make all channels have effectively the same filters as eus channels modulo any rollouts in progress (ie 4.18).

PR #6331, opened 2024-11-28
add a known risk for 4.18.0-ec.3

machine424:
4.18.0-ec.3 shipped the strict CMO config validation https://amd64.ocp.releases.ci.openshift.org/releasestream/4-dev-preview/release/4.18.0-ec.3 and **was built before** 4.17.5 where the "make CMO Upgradeable=false to ensure configurations pass strict validation before the 4.18 upgrade" was shipped https://amd64.ocp.releases.ci.openshift.org/releasestream/4-stable/release/4.17.5

PR #6402, opened 2024-12-06
`CMOSuddenStrictConfigValidation`: Fix regexes

petr-muller:
- ECs and RC numbers are dot-separated
- regexes need to capture [architecture suffixes](https://github.com/openshift/cincinnati-graph-data?tab=readme-ov-file#block-edges)

I was a bit trigger happy when reviewing https://github.com/openshift/cincinnati-graph-data/pull/6331

/cc @wking @machine424


machine424:
Good catch, you bet me to it, I was about to open a PR for the EC/RC dot.
good to know for the suffixes.
thanks
/lgtm

PR #7144, opened 2025-04-25
OCPBUGS-54594: ARM64 nodes to apply a bootloader update at 4.18.11 or later

sdodson:
…update

Reference: https://issues.redhat.com/browse/OCPBUGS-54594

### DIFF
--- a/build-suggestions/4.19.yaml
+++ b/build-suggestions/4.19.yaml
@@ -1,5 +1,5 @@
 default:
-  minor_min: 4.18.6
+  minor_min: 4.18.11
   minor_max: 4.18.9999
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
/lgtm

build-suggestions/4.19.yaml
```diff
@@ -1,5 +1,5 @@
 default:
-  minor_min: 4.18.6
+  minor_min: 4.18.11
```
machine424, line 2: just a note in case this gets reverted.
We also rely on this for https://issues.redhat.com/browse/MON-4205
please reach out to us if you want to revert this.

PR #7193, opened 2025-05-02
OCPNODE-3074: Declare `CRIOLayerCompressionPulls` for 4.18.1..6

petr-muller:
Fixed versions are already available, we are just steering clusters away from the affected ones.


wking (approved):
lgtm

PR #6584, opened 2025-01-15
OTA-1310: Remove stable channel 4.12.z feeders

sdodson:
4.12.z went EUS back in July, however we had a mandate to continue
plumbing them through to stable channels through January and that's
ended now. Since no additional 4.12.z go into stable-4.12, they shouldn't
go into stable-4.13 and stable-4.14 either. Those who continue with 4.12
should use eus channels which are tailored for that use.

PR #7227, opened 2025-05-08
OCPBUGS-55193: channels/stable-*: Remove 4.14.z feeders from stable channels

wking:
Like a56b5405 (#6584), but for 4.14, because [4.14 left its Maintenance phase and entered its EUS phase on 2025-05-01][1]:

```console
$ curl -s 'https://access.redhat.com/product-life-cycles/api/v1/products?name=Openshift+Container+Platform+4' | jq -r '.data[].versions[] | select(.name == "4.14").phases[] | .date + " " + .name'
2023-10-31T00:00:00.000Z General availability
2024-05-27T00:00:00.000Z Full support
2025-05-01T00:00:00.000Z Maintenance support
2025-10-31T00:00:00.000Z Extended update support
2026-10-31T00:00:00.000Z Extended update support Term 2
N/A Extended life phase
```

The most recent 4.14.z is 4.14.51, and it entered stable and related on 354234b97d (#7200), as expected for a release that went GA and entered fast channels before the Maintenance cutoff in 90aff68115 (#7161).  The next 4.14.z to be built will include [OCPBUGS-55193][2], which will default it to the `eus-4.14` channel, so it does not need to enter `stable-*` channels.

[1]: https://access.redhat.com/support/policy/updates/openshift#dates
[2]: https://issues.redhat.com/browse/OCPBUGS-55193

### DIFF
--- a/channels/stable-4.14.yaml
+++ b/channels/stable-4.14.yaml
@@ -1,7 +1,3 @@
-feeder:
-  delay: PT0H
-  filter: 4[.](13|14)[.][0-9].*
-  name: stable
 name: stable-4.14
 versions:
 - 4.12.0
--- a/channels/stable-4.15.yaml
+++ b/channels/stable-4.15.yaml
@@ -1,6 +1,6 @@
 feeder:
   delay: PT0H
-  filter: 4[.](14|15)[.][0-9].*
+  filter: 4[.]15[.][0-9].*
   name: stable
 name: stable-4.15
 versions:
--- a/channels/stable-4.16.yaml
+++ b/channels/stable-4.16.yaml
@@ -1,6 +1,6 @@
 feeder:
   delay: PT0H
-  filter: 4[.](14|15|16)[.][0-9].*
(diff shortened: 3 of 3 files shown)


wking:
I've manually twiddled the `jira/valid-bug` and `jira/invalid-bug` labels, to allow attaching this to OCPBUGS-55193.  The change doesn't ship as part of OCP, so QE will not care about the attachment, but the Jira linking makes the bigger picture more accessible to anyone looking at the bug and wondering why the installer change was made.

sdodson:
/hold
I think we want to wait for 4.14.51 to enter stable-4.15 and stable-4.16 which will mean waiting until 4.16.40 moves to stable.

sdodson:
/lgtm
though

sdodson:
/hold cancel
#7292 cleared things up earlier than normal

sdodson:
/hold
actually lets wait until osus bot looks for things i may have missed

PR #7228, opened 2025-05-08
OCPNODE-3245: Declare `KubeletStartFailingFromRestoreconTimeout` for 4.18.0..12

haircommander:
I'm not totally clear if this is the correct way to do this, and I'm not positive we'll go this route, but beginning the process to indicate we may need it.

### DIFF
--- a/blocked-edges/4.18.0-KubeletStartFailingFromRestoreconTimeout.yaml
+++ b/blocked-edges/4.18.0-KubeletStartFailingFromRestoreconTimeout.yaml
@@ -1,0 +1,8 @@
+to: 4.18.0
+from: 4[.]17[.].*
+url: https://issues.redhat.com/browse/OCPNODE-3245
+name: KubeletStartFailingFromRestoreconTimeout
+message: |-
+  In some cases, a SELinux restorecon run as a prerequisite of kubelet.service was failing.
+matchingRules:
+- type: Always
--- a/blocked-edges/4.18.1-KubeletStartFailingFromRestoreconTimeout.yaml
+++ b/blocked-edges/4.18.1-KubeletStartFailingFromRestoreconTimeout.yaml
@@ -1,0 +1,8 @@
+to: 4.18.1
+from: 4[.]17[.].*
+url: https://issues.redhat.com/browse/OCPNODE-3245
+name: KubeletStartFailingFromRestoreconTimeout
+message: |-
+  In some cases, a SELinux restorecon run as a prerequisite of kubelet.service was failing.
+matchingRules:
+- type: Always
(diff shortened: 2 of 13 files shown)


blocked-edges/4.18.0-MCORestoreCon.yaml
```diff
@@ -0,0 +1,9 @@
+to: 4.18.0
+from: 4[.]17[.].*
+fixedIn: 4.18.13
```
wking, line 3: This only needs to be defined in the most-recent exposed patch release, so if you keep it in the 4.18.12 file, you can drop it from all the others.

blocked-edges/4.18.0-MCORestoreCon.yaml
```diff
@@ -0,0 +1,9 @@
...
+url: https://issues.redhat.com//browse/OCPBUGS-54765
```
wking, line 4: This can point at https://issues.redhat.com/browse/OCPNODE-3245 , which is getting set up with a customer-facing description focused on helping cluster-admins weigh the risk.  The bug is focused on helping developers and QE build and test a fix, so it has content that's not immediately relevant to update-risk-weighing.

blocked-edges/4.18.0-MCORestoreCon.yaml
haircommander: done!

blocked-edges/4.18.0-MCORestoreCon.yaml
haircommander: done!

blocked-edges/4.18.0-KubeletStartFailingFromRestoreconTimeout.yaml
```diff
@@ -0,0 +1,8 @@
+to: 4.18.0
+from: 4[.]17[.].*
+url: https://issues.redhat.com/browse/OCPNODE-3245
+name: KubeletStartFailingFromRestoreconTimeout
+message: |-
+  In some cases, a SELinux restorecon run as a prerequisite of kubelet.service was failing. Now, the restorecon is more targeted.
```
wking, line 6: maybe we should drop "Now, the restorecon is more targeted", because "Now" isn't all that clear?  Folks who are wondering about the fix can click through to OCPNODE-3245 and see  more context there, and folks who see 4.18.13 recommended without mentioning this risk won't be reading this message.

petr-muller (approved):
Content looks good so LGTM (and I really appreciate @haircommander crafting the PR!)

I have some concerns about timing ([slack](https://redhat-internal.slack.com/archives/CK1AE4ZCK/p1747059240027999?thread_ts=1746693051.806679&cid=CK1AE4ZCK)) so I'm holding for now.

/hold

sdodson:
/hold cancel
Causes nodes to go unready requiring per host intervention to return to service and it probably won't be particularly clear to anyone who runs into this how to fix it. OCPNODE-3245 has been updated with remediation steps in case anyone runs into the problem and potential more polished KCS is on the way.

PR #7237, opened 2025-05-09
OCPNODE-3119: Restrict `from:` on `MCOContainerRuntimeConfigStaleFinalizer`

petr-muller:
Only updates from any 4.16 and 4.17 below 4.17.10 are affected, no need to bother 4.17 clusters that either already or never encountered the bug.


hongkailiu (approved):
/lgtm

PR #7356, opened 2025-05-29
OCPBUGS-55638: OCPBUGS-55637: raise minor_min to 4.18.16

djoshy:
This includes admin-acks that handle [1] and [2], which we want cluster admins to evaluate before they update to 4.19.

[1]: https://issues.redhat.com//browse/OCPBUGS-55638
[2]: https://issues.redhat.com//browse/OCPBUGS-55637

### DIFF
--- a/build-suggestions/4.19.yaml
+++ b/build-suggestions/4.19.yaml
@@ -1,5 +1,5 @@
 default:
-  minor_min: 4.18.13
+  minor_min: 4.18.16
   minor_max: 4.18.9999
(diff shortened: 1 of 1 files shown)


wking (approved):
/lgtm

And I manually twiddled the `jira/invalid-bug` and `jira/valid-bug` labels, because graph-data only has a dev-branch (so we don't have `release-4.y` branches).  And we aren't part of what OCP ships in Errata; we're just talking about those bugs in this pull.

PR #7380, opened 2025-06-02
block edges from all 4.18 to 4.19 with vsphere provider storage issues

PratikMahajan:
Clusters running on the vSphere platform with the vSphere CSI driver mounting VSAN Files volumes whenever VSAN is not fully patched and up to date.
Nodes cannot mount NFS volumes from VSAN files provider.
Blocking all the affected edges

### DIFF
--- a/blocked-edges/4.19.0-VSphereStorageMountIssues.yaml
+++ b/blocked-edges/4.19.0-VSphereStorageMountIssues.yaml
@@ -1,0 +1,12 @@
+to: 4.19.0
+from: 4[.]18[.].*
+url: https://issues.redhat.com/browse/STOR-2486
+name: VSphereStorageMountIssues
+message: vSphere customers using vSAN file volumes can't mount vSphere shared volumes and NFS volumes which server do not set NFS4ERR_ATTRNOTSUPP
+matchingRules:
+  - type: PromQL
+    promql:
+      promql: |
+        group(cluster_infrastructure_provider{type=~"vSphere|None"})
+        or
+        0 * group(cluster_infrastructure_provider)
--- a/blocked-edges/4.19.0-ec.0-VSphereStorageMountIssues.yaml
+++ b/blocked-edges/4.19.0-ec.0-VSphereStorageMountIssues.yaml
@@ -1,0 +1,12 @@
+to: 4.19.0-ec.0
+from: 4[.]18[.].*
+url: https://issues.redhat.com/browse/STOR-2486
+name: VSphereStorageMountIssues
+message: vSphere customers using vSAN file volumes can't mount vSphere shared volumes and NFS volumes which server do not set NFS4ERR_ATTRNOTSUPP
+matchingRules:
+  - type: PromQL
+    promql:
+      promql: |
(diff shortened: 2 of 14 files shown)


blocked-edges/4.19.0-ec.1-VsphereStorageMountIssues.yaml
```diff
@@ -0,0 +1,12 @@
+to: 4.19.0-ec.1
+from: 4[.]18[.].*
+url: https://issues.redhat.com/browse/STOR-2486
+name: VsphereStorageMountIssues
+message: Vsphere customers using vSAN file volumes can't mount vSphere shared volumes and nfs volumes which server do not set NFS4ERR_ATTRNOTSUPP
+matchingRules:
+  - type: PromQL
+    promql:
+      promql: |
+        group(cluster_infrastructure_provider{type=~"VSphere|None"})
```
gnufied, line 10: Why does this have `None` also btw?

blocked-edges/4.19.0-ec.0-VsphereStorageMountIssues.yaml
```diff
@@ -0,0 +1,12 @@
+to: 4.19.0-ec.0
+from: 4[.]18[.].*
+url: https://issues.redhat.com/browse/STOR-2486
+name: VsphereStorageMountIssues
+message: Vsphere customers using vSAN file volumes can't mount vSphere shared volumes and nfs volumes which server do not set NFS4ERR_ATTRNOTSUPP
```
wking, line 5: Nit: "Vsphere" -> "VSphere" in the `name` (to match `VSphereHW14CrossNodeNetworkingError` and `VSphereNodeNameChanges`) and "vSphere" in the `message`.  Having both `Vsphere` and `vSphere` in the `message` seems confusing, and [Broadcom seems to prefer "vSphere"][1], even as the opening word in a sentence or title:

image


[1]: https://techdocs.broadcom.com/us/en/vmware-cis/vsphere.html

wking: Another nit: [NFS is an acronym][1], so you probably want "nfs" -> "NFS" in this `message`.  I'm also not seeing any NFS discussion in [STOR-2486][2], just the mention in the title.  Maybe we don't need to talk about NFS volumes at all, and it's just the vSAN volumes?  Maybe NFS volumes and vSAN volumes are the same thing?  I'm not sure.

[1]: https://en.wikipedia.org/wiki/Network_File_System
[2]: https://issues.redhat.com/browse/STOR-2486

PratikMahajan: just took the words from the ticket

blocked-edges/4.19.0-ec.1-VsphereStorageMountIssues.yaml
gnufied: It should only apply to vSphere platform. If someone is bringing their own CSI cloudprovider and CSI driver, they may still be affected but IMO blocking metal etc seems wrong.

PR #7785, opened 2025-08-09
channels/candidate-4.19: Add 4.19.8

wking:
It was tombstoned in c97f27642d (#7780).  But promote it to `candidate-4.19` anyway, because there are [bugfixes to some components since 4.19.7][1], and folks might want to use those signed-but-not-GA images.  For example, in HyperShift, where a HostedCluster update to 4.19.8 doesn't bring in the kernel regression, and NodePools can be left on releases with happier kernels.

[1]: https://amd64.ocp.releases.ci.openshift.org/releasestream/4-stable/release/4.19.8


wking:
/hold

I need to check some risks like `AROMissingInternalLBSAN`...

PR #8038, opened 2025-09-26
CONSOLE-4762: `ConsoleCrashOnMissingPlugin`: Declare on 4.19.(7-13)

petr-muller:
If a Console Operator configuration resource references a plugin name for which there is no corresponding ConsolePlugin resource, the Console may crashloop after the cluster is updated to an affected version.


petr-muller:
Testing showed the original PromQL idea will unfortunately not work, so I turned this into `type: Always` and propose to merge it once a fixed version is available.

/hold

PR #8062, opened 2025-10-01
MCO-1890: blocked-edges/4.19.1[23]-MachineConfigNodesV1AlphaControlPlaneLabels: Declare new risk

wking:
[Fixed in 4.19.14][1].  For the PromQL, the logic is:

* If you have any modern-role `control-plane` Nodes, you are not exposed (0), otherwise...
* ... if you have legacy-role `master` Nodes, you are exposed (1), otherwise...
* ... if the `kube_node_role` metric is working at all, you are not exposed (0), otherwise...
* ... we don't know if you are exposed or not (no response metric).

I'm preserving the `role` label in the output results to make it easier to understand how the PromQL functions if folks run it by hand.

[1]: https://issues.redhat.com/browse/OCPBUGS-62114?focusedId=28155286&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-28155286

### DIFF
--- a/blocked-edges/4.19.12-MachineConfigNodesV1AlphaControlPlaneLabels.yaml
+++ b/blocked-edges/4.19.12-MachineConfigNodesV1AlphaControlPlaneLabels.yaml
@@ -1,0 +1,14 @@
+to: 4.19.12
+from: .*
+url: https://issues.redhat.com/browse/MCO-1890
+name: MachineConfigNodesV1AlphaControlPlaneLabels
+message: Standalone clusters born in 4.11 or earlier whose control-plane nodes lack the control-plane role may need that role added to update to the target release.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      0 * group by (role) (kube_node_role{_id="",role="control-plane"})
+      or on ()
+      1 * group by (role) (kube_node_role{_id="",role="master"})
+      or on ()
+      0 * topk(1, count by (role) (kube_node_role{_id=""}))
--- a/blocked-edges/4.19.13-MachineConfigNodesV1AlphaControlPlaneLabels.yaml
+++ b/blocked-edges/4.19.13-MachineConfigNodesV1AlphaControlPlaneLabels.yaml
@@ -1,0 +1,15 @@
+to: 4.19.13
+from: .*
+fixedIn: 4.19.14
+url: https://issues.redhat.com/browse/MCO-1890
+name: MachineConfigNodesV1AlphaControlPlaneLabels
+message: Standalone clusters born in 4.11 or earlier whose control-plane nodes lack the control-plane role may need that role added to update to the target release.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      0 * group by (role) (kube_node_role{_id="",role="control-plane"})
+      or on ()
+      1 * group by (role) (kube_node_role{_id="",role="master"})
+      or on ()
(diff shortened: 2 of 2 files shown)


blocked-edges/4.19.12-MachineConfigNodesV1AlphaControlPlaneLabels.yaml
```diff
@@ -0,0 +1,14 @@
...
+message: Standalone clusters born in 4.11 or earlier whose control-plane nodes lack the control-plane role may need that role added to update to the target release.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      0 * count by (role) (kube_node_role{_id="",role="control-plane"})
+      or on ()
+      1 * count by (role) (kube_node_role{_id="",role="master"})
+      or on ()
+      0 * topk(1, count by (role) (_id="",kube_node_role))
```
DavidHurta, line 14: image

:thinking:

wking:
Testing the PromQL, a modern standalone cluster knows it isn't exposed:

image

If I tweak the PromQL to use a non-existant role as the "modern" one to simulate an old cluster that hasn't rolled its control-plane nodes, it thinks it is exposed:

image

wking:
oops, but I need 1, not 3...

oarribas:
On the other hand, can the URL to the KCS describing the issue and with the workaround be added? It will be probably more useful for customers than the link to MCO-1890.


[1] https://access.redhat.com/solutions/7131832

wking:
> On the other hand, can the URL to the KCS describing the issue and with the workaround be added...

The KCS seems to be aimed at the "so, you got bit.  Here's what to do..." situation.  The Jira impact statement is aimed at the "so, you're considering an update into an exposed release.  Here's how to figure out if you're actually exposed, and if you are, what that exposure looks like, and what you can do about it" situation.  I'd rather link the Jira from this risk, because folks who see the message are in that considering-whether-to-update situation.  The mitigation section of the Jira should definitely link the KCS though, to help folks who hear about the risk after getting bit have that "you got bit..." focused text easily available.

DavidHurta (approved):
/lgtm
/hold

Holding as there is a discussion. Feel free to unhold when applicable.

wking:
> If there are no master and not control-plane (last case), will OCPBUGS-62073 not happens?

I have [four cases][1], and it's not clear which you're talking about.  Once case where there are neither `master` nor `control-plane` would be HyperShift.  I don't have any HyperShift clusters handy to test in, but I'm also not aware of any HyperShift deployments that have functional PromQL update-risk evaluation, so 🤷.  But ideally HyperShift clusters would be able to see "oh, I'm not exposed", because there was no HyperShift back in in 4.11, and [the CronJob doesn't run on HyperShift][2] anyway.

However, there's also the neither `master` nor `control-plane` because the `kube_node_role` is busted" case, and in that case I want to return no results, so the CVO sets `Recommended=Unknown` in ClusterVersion `status`.  The cluster admin would have to self-assess their exposure.

[1]: https://github.com/openshift/cincinnati-graph-data/pull/8062#issue-3474435741
[2]: https://github.com/openshift/machine-config-operator/blob/2b5b7ce2126ece73a249e6aa829b53fcb3320dbe/install/0000_80_machine-config_00_v1alpha1-mcn-cleanup-job.yaml#L9-L10

oarribas:
Thanks for the clarifications

PR #9286, opened 2026-04-29
Drop 4.15.63 tombstoned release from candidate channels

germanparente:


germanparente:
@wking I know this is not urgent but don't hesitate to approve this as possible

channels/candidate-4.15.yaml
```diff
@@ -153,4 +153,3 @@ versions:
 - 4.15.60
 - 4.15.61
 - 4.15.62
-- 4.15.63
```
wking, line 156: We can tombstone releases, but we don't drop them from channels.  Dropping would make it harder for folks on that release in that channel to update out to a fixed release.

rhdmalone: hi @wking 
Does the same also apply to channels/candidate-4.16.yaml ?
i.e. do not remove the row from channels/candidate-4.16.yaml also?
Thanks

wking: yup, applies to all channels

germanparente:
Thanks @wking I have updated the PR to tombstone 4.15.63.

so, the traces of 4.15.63 are still in 4.15 / 4.16 channels.

rhdmalone:
@wking 
if possible, could you approve this PR as it tidies up tombstoning of 4.15.63 from last week please ?
Thanks very much

internal-channels/candidate.yaml
```diff
@@ -1773,12 +1775,11 @@ versions:
 - 4.19.29
 - 4.21.12
 - 4.16.60
-- 4.15.63
```
wking, line 1776: You're still dropping this entry.  Doesn't really matter, because it's currently only an internal channel.  But all you should be doing when tombstoning is adding the `tombstones` entry up around line 115.

germanparente: Thanks @wking I will fix this asap. 
I know that candidate channel is not really very importante since it's not even supported. Thanks for reminding me of that.

PR #9303, opened 2026-04-30
OTA-1988: hack/release-ga: Check openshift-eng/ga-regression-reports

wking:
As described in [the new repository][1], an approved regression report is now required before a new OpenShift release can go Generally Available.  We might grow presubmit coverage for centralized enforcement later, but for now, update the script that maintainers run locally to check for a report.

[1]: https://github.com/openshift-eng/ga-regression-reports

### DIFF
--- a/hack/release-ga.sh
+++ b/hack/release-ga.sh
@@ -37,6 +37,9 @@ then
 	exit 1
 fi
 
+REGRESSION_REPORT="https://github.com/openshift-eng/ga-regression-reports/blob/main/openshift/${MAJOR}.${MINOR}-ga-regression-report.md"
+curl -s --fail "${REGRESSION_REPORT}" || (echo "failed to retrieve regression report required for GA: ${REGRESSION_REPORT}" >&2; exit 1)
+
 if test "$((MINOR % 2))" -eq 0
 then
 	# 4.even get extended update support: https://issues.redhat.com/browse/OTA-1599


dgoodwin:
/lgtm

ANd thanks!

hack/release-ga.sh
```diff
@@ -37,6 +37,9 @@ then
 	exit 1
 fi
 
+REGRESSION_REPORT="https://github.com/openshift-eng/ga-regression-reports/blob/main/openshift/${MAJOR}.${MINOR}-ga-regression-report.md"
+curl -s --fail "${REGRESSION_REPORT}" || (echo "failed to retrieve regression report required for GA: ${REGRESSION_REPORT}" >&2; exit 1)
```
hongkailiu, line 41: nit: We could add `-o /dev/null` if we care only about the existence of the file, not its content.
```suggestion
curl -s --fail -o /dev/null "${REGRESSION_REPORT}" || (echo "failed to retrieve regression report required for GA: ${REGRESSION_REPORT}" >&2; exit 1)
```

wking:
I'm not sure where the right place for this guard is, but it should be before ART pushes the Errata.  This pull request is bumping a script that comes after the Errata push, and witholding a release from `fast-*` channels despite a public Errata isn't helpful (Red Hat has already declared the release supported).  Closing.

PR #9610, opened 2026-06-09
NO-ISSUE: *: 4.22.0 goes generally available

wking:
Run:

```console
$ hack/release-ga.sh 4.22
```


wking:
self-LGTM, because this is just automated output from a version-controlled script

PR #9640, opened 2026-06-11
OCPNODE-4579: Declare the S390xContainerDataFailure risk

DavidHurta:


DavidHurta:
/hold

verifying

DavidHurta:
The build11 cluster is a multi cluster with a s390x node:

image

The build12 is a multi cluster with no s390x nodes:

image

The ota-stage cluster is a single-arch amd64 cluster

image

blocked-edges/4.22.0-ec.3-S390xContainerDataFailure.yaml
```diff
@@ -0,0 +1,15 @@
+to: 4.22.0-ec.3
+from: ^4[.](21[.].*|22[.]0-ec[.][0-2])[+].*$
```
DavidHurta, line 2: I wanted to leverage the arch suffix for something like `(s390x|multi)`, but I am not 100% sure whether the relevant logic processes multiarch versions the same as single arch versions.

The graph builder does output versions with the `+multi` suffix. However, we can explore this later and refine the risk if needed.

wking (approved):
/lgtm

PR #9718, opened 2026-06-19
CONSOLE-5337: blocked-edges: Add ControlPlaneStatusGreyIcon risk for 4.20.23-24

hongkailiu:
Add upgrade risk for [CONSOLE-5337](https://redhat.atlassian.net/browse/CONSOLE-5337) affecting 4.20.23 and 4.20.24 where the control plane status indicator remains grey and never shows green, even when cluster operators are healthy. Fixed in 4.20.25.

Affects all clusters except those with external control plane topology (HyperShift).

PromQL pattern is based on HyperShiftRedundantRouter[1] but inverted:
- HyperShiftRedundantRouter targets HyperShift clusters (invoker="hypershift")
- This risk excludes HyperShift clusters by zeroing them out (0 * ...)
- The "or on ()" clause then returns all non-HyperShift clusters
- Difference: HyperShiftRedundantRouter affects only HyperShift, this affects only standalone (non-HyperShift) clusters

[1] https://github.com/openshift/cincinnati-graph-data/blob/master/blocked-edges/4.19.18-HyperShiftRedundantRouter.yaml

```console
$ diff -Naupr blocked-edges/4.20.23-ControlPlaneStatusGreyIcon.yaml blocked-edges/4.20.24-ControlPlaneStatusGreyIcon.yaml
--- blocked-edges/4.20.23-ControlPlaneStatusGreyIcon.yaml       2026-06-19 11:29:45
+++ blocked-edges/4.20.24-ControlPlaneStatusGreyIcon.yaml       2026-06-19 11:29:46
@@ -1,5 +1,6 @@
-to: 4.20.23
+to: 4.20.24
 from: .*
+fixedIn: 4.20.25
 url: https://redhat.atlassian.net/browse/CONSOLE-5337
 name: ControlPlaneStatusGreyIcon
 message: |
```

/hold

verifying the PromQL

### DIFF
--- a/blocked-edges/4.20.23-ControlPlaneStatusGreyIcon.yaml
+++ b/blocked-edges/4.20.23-ControlPlaneStatusGreyIcon.yaml
@@ -1,0 +1,13 @@
+to: 4.20.23
+from: .*
+url: https://redhat.atlassian.net/browse/CONSOLE-5337
+name: ControlPlaneStatusGreyIcon
+message: |
+  Control plane status indicator remains grey and never shows green, even when cluster operators are healthy.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      0 * group by (invoker) (cluster_installer{_id="",invoker="hypershift"})
+      or on ()
+      group by (invoker) (cluster_installer{_id=""})
--- a/blocked-edges/4.20.24-ControlPlaneStatusGreyIcon.yaml
+++ b/blocked-edges/4.20.24-ControlPlaneStatusGreyIcon.yaml
@@ -1,0 +1,14 @@
+to: 4.20.24
+from: .*
+fixedIn: 4.20.25
+url: https://redhat.atlassian.net/browse/CONSOLE-5337
+name: ControlPlaneStatusGreyIcon
+message: |
+  Control plane status indicator remains grey and never shows green, even when cluster operators are healthy.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      0 * group by (invoker) (cluster_installer{_id="",invoker="hypershift"})
+      or on ()
+      group by (invoker) (cluster_installer{_id=""})


hongkailiu:
1 on `ota-stage` as it is a HA cluster.

```
$ oc --kubeconfig /tmp/ota-stage.c get infrastructure cluster -o yaml | yq -r .status.controlPlaneTopology
HighlyAvailable
```

Screenshot 2026-06-19 at 11 22 47


But a cluster bot with `launch 4.22` ([logs](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-hypershift-hosted/2067991339038937088)) shows

Screenshot 2026-06-19 at 11 51 52

How did it work for https://github.com/openshift/cincinnati-graph-data/blob/master/blocked-edges/4.19.18-HyperShiftRedundantRouter.yaml ?

The same for metrics `cluster_infrastructure_provider`.

Maybe we could use `cluster_infrastructure_topology OR on() vector(0)` to return 1 on HA and 0 on HyperShift hosted clusters?

Or maybe just use `type: Always` as suggested in https://redhat.atlassian.net/browse/CONSOLE-5337?focusedCommentId=17285358?

/hold

wking (approved):
Re: [HyperShift evaluation][1], what you have is our current recommendation.  The check you ran seems to have been on the Prometheus/Thanos instance hosted in the compute Nodes, and that's not the PromQL store that the cluster-version operator will be using.  For example, see the openshift/hypershift#7399 series setting up the ROSA HCP clusters to evaluate PromQL against the management cluster's RHOBS store.

Does the `cluster_installer` metric work in the RHOBS store?  I don't have access to a ROSA management cluster to check.  But it's the pattern we've used for HyperShift so far, and if it fails to evaluate, the CVO will count the cluster as exposed and bubble the risk up in ClusterVersion `status` (from where it will be bubbled up further through HostedControlPlane `status` to HostedCluster `status` and beyond).  If ROSA HCP folks notice their clusters failing to evaluate, they can work with us to refine the PromQL.  But I think we can safely merge this as it stands, regardless of whether or not the PromQL successfully evaluates "I'm not exposed" on ROSA HCP or other HyperShift deployments, because failing closed (and possibly triggering some engagement with folks maintaining HyperShift clusters to work out what the PromQL should actually be) seems like less of a risk than delaying declaration (and leaving some standalone cluster admins unaware of a risk that might bite their cluster on update).

/lgtm
/hold cancel

[1]: https://github.com/openshift/cincinnati-graph-data/pull/9718#issuecomment-4752972335

PR #9960, opened 2026-07-13
hack: Improve risk extension guidance in stabilization-changes

TheRealJon:
The previous message told users to "declare a fix version or extend the risk" without explaining how to determine which action to take or how to do either one. Replace with step-by-step instructions that direct users to check the issue URL, then either set fixedIn or copy the blocked-edges entry with an updated "to" field.


wking (approved):
/lgtm

PR #10046, opened 2026-07-21
OCPBUGS-83862: Add conditional update risk for 4.21.0-25 from 4.20.z < 4.20.25

shahsahil264:
## Summary

Add `type: Always` conditional update risk for all existing 4.21.z GA releases (4.21.0 through 4.21.25) warning clusters upgrading from 4.20.24 or earlier about the Auto Node Sizing issue.

## Context

PR #9494 raises `minor_min` from 4.20.15 to 4.20.25 for future 4.21.z builds. However, all existing 4.21.z releases were built with the old `minor_min`, so clusters on 4.20.0-4.20.24 can still see upgrade paths to them.

Per [Scott Dodson](https://redhat-internal.slack.com/archives/CEGKQ43CP/p1784554046680519) and [W. Trevor King](https://redhat-internal.slack.com/archives/CEGKQ43CP/p1784590741323619), these existing versions need a conditional update risk to warn affected clusters.

## Bug

[OCPBUGS-83862](https://redhat.atlassian.net/browse/OCPBUGS-83862): During upgrade from pre-4.20.25, `auto-sizing-disabled` MachineConfigs override KubeletConfig autosizing settings on custom MachineConfigPool nodes. Fixed in 4.20.25 via [openshift/machine-config-operator#5885](https://github.com/openshift/machine-config-operator/pull/5885).

## Risk details

- **from**: `^4[.]20[.](1?[0-9]|2[0-4])[+].*$` — matches 4.20.0 through 4.20.24
- **type**: Always — affects all clusters upgrading from the unpatched version range
- **26 files**: one per GA 4.21.z release (4.21.0 through 4.21.25)

Companion to #9494.

PR #9494, opened 2026-05-20
Set minimum version of 4.20 required to upgrade to 4.21

aksjadha:
Fixes: Related to fix #[OCPBUGS-83862](https://redhat.atlassian.net/browse/OCPBUGS-83862) to be applied before upgrading to 4.21

This change introduces a gate to block 4.20 clusters from upgrading to 4.21 until the prerequisite patch for Auto Node Sizing has been applied.

Requirement: The necessary 4.20 patch https://github.com/openshift/machine-config-operator/pull/5885 introduces a specific MachineConfig to enforce this disabled state for master and worker mcp. This is fixed in 4.20.25.

Goal: This PR ensures upgrade safety by confirming the presence of that MachineConfig before proceeding to 4.21.

PR #10047, opened 2026-07-21
"MCO-2489: Raise minor_min to 4.20.32 for 4.21"

shahsahil264:
## Summary

Combined PR: raises `minor_min` from 4.20.15 to 4.20.25 for future 4.21.z builds, and adds conditional update risk for all existing 4.21.z GA releases (4.21.0 through 4.21.25).

Cherry-picks the build-suggestions change from #9494 (by @aksjadha) and adds the companion blocked-edge files requested by @sdodson and @wking.

Replaces #9494 and #10046.

---

Update by @hongkailiu 

* Claim AutoSizingReservedDisabledSilently for upgrade
  * 4.19 -> 4.20.z (z<32)
  * 4.20 -> 4.21.*: After merging, we will have to extend the risk to all future releases for 4.21 as the fix is not shipped to 4.21 (and there is no plan of doing it).

* Raise `minor_min` to 4.20.32 for 4.21. It helps reduce the number of 4.20 clusters that carries over the issue to 4.21.

## Bug

[OCPBUGS-83862](https://redhat.atlassian.net/browse/OCPBUGS-83862): During upgrade from pre-4.20.25, `auto-sizing-disabled` MachineConfigs override KubeletConfig autosizing settings on custom MachineConfigPool nodes. Fixed in 4.20.25 via [openshift/machine-config-operator#5885](https://github.com/openshift/machine-config-operator/pull/5885).

## Risk details

- **from**: `^4[.]20[.](1?[0-9]|2[0-4])[+].*$` — matches 4.20.0 through 4.20.24
- **PromQL**: Scoped to clusters with KubeletConfigs via `apiserver_storage_objects` (per Trevor's recommendation — 26k clusters without KubeletConfigs won't see the warning)
- **26 blocked-edge files**: one per GA 4.21.z release (4.21.0 through 4.21.25)


build-suggestions/4.21.yaml
```diff
@@ -1,5 +1,5 @@
 default:
-  minor_min: 4.20.15
+  minor_min: 4.20.25
```
wking: Poking into the implementation, I don't think 4.20.25 is sufficient, and I've reported the issue as [OCPBUGS-99650](https://redhat.atlassian.net/browse/OCPBUGS-99650) .  I don't think we should be raising the floor until we know when it's actually fixed.  And given the low impact rate and the number of clusters who would be updated by a floor-raise, it might be too late in the 4.20.z lifecycle to declare at all?

DavidHurta: >  I don't think we should be raising the floor until we know when it's actually fixed.

Agree.

> And given the low impact rate and the number of clusters who would be updated by a floor-raise, it might be too late in the 4.20.z lifecycle to declare at all?

Yes, it's a larger ask. Definitely something to ponder about a little bit before introducing.

blocked-edges/4.21.9-AutoSizingDisabledMachineConfigOverride.yaml
```diff
@@ -0,0 +1,13 @@
+to: 4.21.9
+from: ^4[.]20[.](1?[0-9]|2[0-4])[+].*$
+url: https://redhat.atlassian.net/browse/OCPBUGS-83862
```
DavidHurta: Currently, the https://redhat.atlassian.net/browse/OCPBUGS-83862 has a security level assigned.

This will hinder the `url`'s purpose, making it less useful for folks interested in more information. Meaning, we can either create a new impact statement jira with public-facing information, or the bug can be evaluated to see whether the assigned security level is justified or whether it can be open to the public (assuming the bug's description is helpful for the conditional risk).

DavidHurta: Currently, not that actionable because of https://github.com/openshift/cincinnati-graph-data/pull/10047#discussion_r3641645077, but it is something to address before merging if we decide to merge the PR.

hongkailiu:
/hold

We will wait until the fix is available for main and all previous releases to 4.20.
Then probably declare the risks without bumping `minor_min` for 4.21.
The relevant [discussion](https://redhat-internal.slack.com/archives/C0B2C9VAHLK/p1785788663519909?thread_ts=1785766022.622459&cid=C0B2C9VAHLK) on slack.

blocked-edges/4.21.9-AutoSizingDisabledMachineConfigOverride.yaml
hongkailiu: 83862 seems open to everyone now.

I will probably still create "a new impact statement jira" to save the energy of our users on jumping between two bugs [OCPBUGS-83862](https://redhat.atlassian.net/browse/OCPBUGS-83862) and [OCPBUGS-99650](https://redhat.atlassian.net/browse/OCPBUGS-99650).

I will update the link soon.

hongkailiu: https://redhat.atlassian.net/browse/MCO-2489

hongkailiu:
/retitle "[MCO-2489](https://redhat.atlassian.net/browse/MCO-2489): Raise minor_min to 4.20.32 for 4.21 and block upgrades for AutoSizingReservedDisabledSilently"

hongkailiu:
Currently, the pull both bumps `minor_min` and claim the risk.

If we claim the risk, we have to claim the risk for each 4.21 future releases. It is because there is no fix version for 4.21 clusters. We have to fix it manually. It cannot be done by a patch to 4.21 because it is technically difficult.

It is not cool but I do not have a better solution for it.

Another choice is we do not claim the risk at all. Maybe there are not many of clusters that enables `autoSizingReserved` on custom MCPs. 4.20 and 4.21 are there for a while, do we have complains of users about it?

I believe in any case we want minor_min: 4.20.32 to reduce the number of carryovers.

hongkailiu:
/retitle "[MCO-2489](https://redhat.atlassian.net/browse/MCO-2489): Raise minor_min to 4.20.32 for 4.21"

PR #10082, opened 2026-07-22
OTA-646: Add GitHub App authentication support to stabilization bot

shahsahil264:
## Summary

Adds GitHub App authentication as an alternative to PAT-based auth in `hack/stabilization-changes.py`. This is the code change needed to migrate `openshift-ota-bot` from a regular GitHub user account to a GitHub App, which fixes the Prow `trusted_apps` issue (Prow requires the `[bot]` suffix that only GitHub Apps have).

## Changes

- New flags: `--github-app-id` and `--github-app-private-key-file` (with `GITHUB_APP_ID` and `GITHUB_APP_PRIVATE_KEY_FILE` env var fallbacks)
- When app credentials are provided, uses PyGithub's `AppAuth` + `GithubIntegration` for API calls and `x-access-token:<installation_token>` for git push
- Backwards compatible: `--github-token` / `GITHUB_TOKEN` continues to work
- No new dependencies (PyGithub already supports GitHub App auth natively)

## Testing

- Created a dev GitHub App (`openshift-ota-bot-dev`, App ID: 4368317)
- Verified full flow: git push with `x-access-token` + PR creation via API
- PR correctly authored by `openshift-ota-bot-dev[bot]` (the `[bot]` suffix Prow needs)
- Test PR: https://github.com/shahsahil264/cincinnati-graph-data/pull/2 (closed)

## Remaining work (separate tickets)

1. Create production GitHub App and transfer to `openshift-eng` org
2. Install app on `openshift/cincinnati-graph-data`
3. Update PSI deployment secrets
4. Verify Prow `trusted_apps` works with `[bot]` suffix
5. Retire old user account

## Bug

https://redhat.atlassian.net/browse/OTA-646

### DIFF
--- a/hack/stabilization-changes.py
+++ b/hack/stabilization-changes.py
@@ -632,7 +632,8 @@ def notify(message, webhook=None):
 
 
 def promote(version, channel_name, channel_path, subject, body, upstream_github_repo, push_github_repo, github_token, upstream_branch, labels=None, github_app_id=None, github_app_private_key=None):
-    if github_token or github_app_id:
+    use_app_auth = github_app_id and github_app_private_key
+    if github_token or use_app_auth:
         upstream_remote = get_remote(repo=upstream_github_repo)
         subprocess.run(['git', 'fetch', upstream_remote], check=True)
         branch = 'promote-{}-to-{}'.format(version, channel_name)
@@ -659,7 +660,7 @@ def promote(version, channel_name, channel_path, subject, body, upstream_github_
         yaml.safe_dump(data, f, default_flow_style=False)
     message = '{}\n\n{}\n'.format(subject, textwrap.fill(body, width=76))
 
-    if not github_token and not github_app_id:
+    if not github_token and not use_app_auth:
         pull = PullRequest(html_url='data://no-token-so-no-pull')
         return pull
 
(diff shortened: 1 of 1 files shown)


hack/stabilization-changes.py
```diff
@@ -655,31 +660,50 @@ def promote(version, channel_name, channel_path, subject, body, upstream_github_
...
     if github is None:
         raise github_import_error
 
     subprocess.run(['git', 'commit', '--file', '-', channel_path], check=True, encoding='utf-8', input=message)
-    push_uri_with_token = 'https://{}@github.com/{}.git'.format(github_token, push_github_repo)
-    subprocess.run(['git', 'push', '-u', push_uri_with_token, branch], check=True)
 
-    owner = push_github_repo.split('/')[0]
+    if use_app_auth:
+        auth = AppAuth(int(github_app_id), github_app_private_key)
```
wking: Instead of casting to `int` here, can we do that up at argument-input-time?  See [these docs][1] to push it all the way up to the `add_argument` call.

[1]: https://docs.python.org/3/library/argparse.html#type

hack/stabilization-changes.py
```diff
@@ -766,6 +804,15 @@ def main():
 
     args = parser.parse_args()
 
+    github_app_private_key = None
+    if args.github_app_private_key_file:
+        key_path = args.github_app_private_key_file.strip()
+        if not os.path.exists(key_path):
+            _LOGGER.fatal('GitHub App private key file not found: {}'.format(key_path))
+            raise SystemExit(1)
```
wking: I don't think we need to special-case this.  We can drop these three lines, and if the file doesn't exist (or we don't have read access, or whatever), the next line's `open` call will raise an exception that will cause the run to exit non-zero with both the error message and a traceback.

PR #10591, opened 2026-09-08
blocked-edges: add risk for Portworx on 4.22.9

simkam:
Clusters running Portworx may lose storage after upgrading to 4.22.9. The RHCOS kernel (5.14.0-687.35.1.el9_8) shipped with that version is missing from RHEL package repositories, preventing Portworx from building its kernel module.

Spike: https://redhat.atlassian.net/browse/OTA-2119
Bug: https://redhat.atlassian.net/browse/OCPBUGS-121207

### DIFF
--- a/blocked-edges/4.22.9-PortworxKernelModuleBuildFailure.yaml
+++ b/blocked-edges/4.22.9-PortworxKernelModuleBuildFailure.yaml
@@ -1,0 +1,17 @@
+to: 4.22.9
+from: .*
+url: https://redhat.atlassian.net/browse/OTA-2119
+fixedIn: 4.22.10
+name: PortworxKernelModuleBuildFailure
+message: |
+  Clusters running Portworx may have storage unavailable after upgrading to 4.22.9.
+  The RHCOS kernel version shipped with 4.22.9 (5.14.0-687.35.1.el9_8) is not
+  available in RHEL package repositories, preventing Portworx from building its
+  kernel module on upgraded nodes.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      group by (_id, name) (csv_succeeded{_id="",name=~"portworx-operator[.]v.*"})
+      or on (_id)
+      0 * label_replace(group by (_id) (csv_succeeded{_id=""}), "name", "portworx-operator not installed", "", "")


blocked-edges/4.22.9-PortworxKernelModuleBuildFailure.yaml
```diff
@@ -0,0 +1,17 @@
+to: 4.22.9
+from: .*
+url: https://redhat.atlassian.net/browse/OCPBUGS-121207
```
wking: we usually point at the impact statement ticket, in this case [OTA-2119](https://redhat.atlassian.net/browse/OTA-2119), and not at the OCPBUGS.  Reasons include:

* The OCPBUGS is often written from a reporter to the component developer, to help them understand what's broken, so they can work up a fix.  In this case, with [OCPBUGS-121207](https://redhat.atlassian.net/browse/OCPBUGS-121207), I'm not sure what the fix would be.  4.22.10 has already shipped, so there's no change to make in the shipped product, right?   The impact statement, on the other hand, is written to the cluster admin considering an update into an impacted release.  So it can cover things like "[How involved is remediation?](https://github.com/openshift/enhancements/blob/a05b35e6f46f0539ee3fe05f74aabe456802d581/enhancements/update/update-blocker-lifecycle/README.md?plain=1#L110)" that a cluster-admin likely cares about, but which a developer trying to fix a bug likely does not care about.
* OCPBUGS are tied to a specific target release, and one issue impacting multiple branches will have a separate OCPBUGS tracking a backport into each branch.  Impact statements can talk about the issue as it impacts updates, without being pinned to a single target release branch.

blocked-edges/4.22.9-PortworxKernelModuleBuildFailure.yaml
```diff
@@ -0,0 +1,17 @@
...
+message: |
+  Clusters running Portworx may have storage unavailable after upgrading to 4.22.9.
+  The RHCOS kernel version shipped with 4.22.9 (5.14.0-687.35.1.el9_8) is not
+  available in RHEL package repositories, preventing Portworx from building its
+  kernel module on upgraded nodes.
+matchingRules:
+- type: PromQL
+  promql:
+    promql: |
+      group by (_id) (csv_succeeded{_id="",name=~"portworx-operator[.]v.*"})
```
wking: It doesn't impact cluster-version operator functionality, but personally, I find this kind of PromQL easier to debug if we keep `name` around instead of `group` aggregating it away.  [This][1] is a recent example

[1]: https://github.com/openshift/cincinnati-graph-data/blob/26ed576af29b135119740622666fd684695e8a41/blocked-edges/4.21.11-PrecisionTimeProtocolDPLLPins.yaml#L11-L13

blocked-edges/4.22.9-PortworxKernelModuleBuildFailure.yaml
simkam: Thanks! Updated.

blocked-edges/4.22.9-PortworxKernelModuleBuildFailure.yaml
simkam: Thanks! Updated.

PR #10646, opened 2026-09-13
Modified to get the exact path as specified in the graph

PratikUplenchwar:
Modified to get the exact path, with the latest version from which the further path is available.  README is also changed


PratikUplenchwar:
There are count less customer facing this issues: https://redhat.atlassian.net/browse/CPLABS-3530  one in today morning via case: 04535837, so I understand that they are fixing the UI, however what I have done is to get the python cli command line till now, which shows latest path with all the known issues, which we can provide it with the customer.
Let me know your valuable remark on this @DavidHurta TIA

PR #10711, opened 2026-09-18
ART-23264: build-suggestions: min_versions schema bump

wking:
openshift-eng/art-tools@f27aa9c04d (openshift-eng/art-tools#3386) taught ART to consume this new schema, which we need in order to declare 5.1 build suggestions like:

```yaml
min_versions:
- 4.23.0-rc.0
- 5.0.0-rc.0
- 5.1.0-ec.0
```

It also drops all the older properties like `*_max`, `*_block_list`, and the per-architecture config options that we'd never needed to use, and don't expect to need in the future.

I'm also adding 4.23 and 5.1 build suggestions, so future ECs in those streams will have update sources in `previous`.  I'd like to have the `hack/release-`* scripts ported too, but they will take more thinking, so for now, I'm just hand-coding the new files.


hongkailiu:
> I'd like to have the hack/release-* scripts ported too, but they will take more thinking

I was about to ask about it. Thanks.

/lgtm

/hold

Free to cancel when ready.

wking:
I've pushed f216efcca58d6840cd2eb892005c4d20b289c7c1 updating the `release-*` script handling too; details in that commit message.

/hold cancel

wking:
And I've also pushed 94cb368bc and 27cf59d0b to catch up `candidate-4.23` and `candidate-5.1` configuration.

PR #10717, opened 2026-09-21
ROSAENG-67429: Rename and update the outdated sre-approvers alias

joshbranham:
This renames and updates the `sre-approvers` alias to contain active members of ROSA, ARO, and GCP technical staff, while removing users who have left the org/company.


wking (approved):
/lgtm

